PR #15, opened 2023-02-17
Create Panel Details page

### LINKED ISSUE #9: Create solar panel parameter input page
## Description

Build out the page for inputting solar panel parameters. 

## Todos

- [x] create input field components  
- [x] put together form page
- [x] style according to [figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=66%3A881&t=VbR70BBkmE7bxXsb-1) (colors will likely change w/design iteration!)
- [ ] [Required](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required) - use this appropriately

## Useful links

- [figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=66%3A881&t=VbR70BBkmE7bxXsb-1)

## Screenshots
Screen Shot 2023-02-16 at 1 39 52 PM

## Time estimate
tbd when assigned!

ashayp22:
# Pull Request

## Description

Build out the page for inputting solar panel parameters. The page has three inputs: solar panel direction, solar panel angle tilt, and solar panel capacity (optional). The page can be displayed in light-mode and dark-mode, and is designed mobile-first.

Fixes #9

## How Has This Been Tested?

This code has been tested manually by a human being. It has been tested against mobile and desktop screen sizes. It has also been tested by [emulatating the prefers-color-scheme media query in Google Chrome](https://stackoverflow.com/questions/57606960/how-can-i-emulate-prefers-color-scheme-media-query-in-chrome).

See screenshots below:

Screenshot 2023-02-17 at 5 31 44 PM

Screenshot 2023-02-17 at 5 31 44 PM

Screenshot 2023-02-17 at 5 31 32 PM

## Checklist:

- [x] My code follows [OCF's coding style guidelines](https://github.com/openclimatefix/nowcasting/blob/main/coding_style.md)
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked my code and corrected any misspellings


pages/form.tsx
```diff
@@ -0,0 +1,114 @@
+import Head from 'next/head';
+
+const Form = () => {
+  return (
+    <>
+      <Head>
+        <title>Panel Details | Open Climate Fix</title>
+        <meta name="description" content="Generated by Neha and Ashay" />
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
+        <link rel="icon" href="/favicon.ico" />
```
AndrewLester: These three head tags are common across the whole site, so they should be moved to `_app.tsx`. For now, leave the description as `pv-sites-mobile`.

pages/form.tsx
```diff
@@ -0,0 +1,114 @@
...
+      </Head>
+      <main className="bg-white dark:bg-black flex flex-col items-center justify-start px-10 min-h-screen">
```
AndrewLester: Since there's already a `main` element wrapping `<Component ...` inside `_app.tsx`, there will be two mains on the page. Moving these styles up to the `main` inside `_app.tsx` should work, and then also prevent style duplication here. To concat the `inter.className` with these tailwind styles, you can just drop the `inter.className` into a template string.

In short, this instance of `<main ...` can be removed.

pages/form.tsx
```diff
@@ -0,0 +1,114 @@
...
+              placeholder="3000 W"
+              className="mt-3 text-lg text-center bg-gray-50 border border-gray-300 text-gray-900 rounded-lg hover:ring-ocf-yellow hover:border-ocf-yellow block w-28 h-14 p-2.5 dark:bg-ocf-gray-800
+              dark:border-ocf-gray-800
+              dark:placeholder-ocf-gray-900 dark:text-ocf-gray"
+            />
+          </div>
+
+          <div className="flex justify-center items-center w-full mt-24">
+            <button className="bg-ocf-yellow dark:bg-ocf-yellow shadow h-14 w-40 text-center rounded-md font-bold text-xl">
+              NEXT
```
AndrewLester: Since this button's text would be capitalized no matter what is said, I think it makes sense to give this the `uppercase` style, then use standard sentence case on the text for readability.
```suggestion
            <button className="bg-ocf-yellow dark:bg-ocf-yellow shadow h-14 w-40 text-center rounded-md font-bold text-xl uppercase">
              Next
```

AndrewLester (changes requested):
Great job on this, just a few small changes.

I'm also noticing some `className` duplication on the `input`s, which potentially could be simplified with another component declared in the file. I think this would improve the experience if we need to change any of the styles, though it may take a bit more time to appropriately handle different `input` width, and other attributes, via props.

pages/form.tsx
```diff
@@ -0,0 +1,114 @@
...
+              const solarPanelAngelTilt = data.get(
+                'solarPanelAngelTilt'
+              ) as string;
+              const solarPanelCapacity = data.get(
+                'solarPanelCapacity'
+              ) as string;
+
+              console.log(
+                `Direction: ${solarPanelDirection}, Angel Tilt: ${solarPanelAngelTilt}, Capacity: ${solarPanelCapacity}`
+              );
```
ericcccsliu: we should be able to remove this console.log

pages/form.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+              const solarPanelDirection = data.get(
+                'solarPanelDirection'
+              ) as string;
+              const solarPanelAngelTilt = data.get(
+                'solarPanelAngelTilt'
+              ) as string;
+              const solarPanelCapacity = data.get(
+                'solarPanelCapacity'
+              ) as string;
```
ericcccsliu: I think these should be typed as `number`—this would also bring them in line with the types you specified on the `<input>` elements.

pages/form.tsx
```diff
@@ -0,0 +1,114 @@
...
+            e.preventDefault();
+            try {
+              const data = new FormData(e.currentTarget);
+
+              const solarPanelDirection = data.get(
+                'solarPanelDirection'
+              ) as string;
+              const solarPanelAngelTilt = data.get(
+                'solarPanelAngelTilt'
+              ) as string;
```
ericcccsliu: typo in "angle"

pages/form.tsx
```diff
@@ -0,0 +1,114 @@
...
+          <div className="my-8">
+            <h2 className="font-medium text-lg dark:text-ocf-gray">
+              Solar panel angle tilt
+            </h2>
+            <p className="text-ocf-gray-900 text-xs mt-1 dark:text-ocf-gray-800">
+              (Degrees above the horizontal)
+            </p>
+            <input
+              required
+              name="solarPanelAngelTilt"
```
ericcccsliu: angle is spelled wrong here too

pages/form.tsx
```diff
@@ -0,0 +1,114 @@
...
+            </div>
+            <input
+              name="solarPanelCapacity"
+              type="number"
+              placeholder="3000 W"
+              className="mt-3 text-lg text-center bg-gray-50 border border-gray-300 text-gray-900 rounded-lg hover:ring-ocf-yellow hover:border-ocf-yellow block w-28 h-14 p-2.5 dark:bg-ocf-gray-800
+              dark:border-ocf-gray-800
+              dark:placeholder-ocf-gray-900 dark:text-ocf-gray"
+            />
+          </div>
```
ericcccsliu: I'd love to get some extra validation on the angle and capacity input elements—seems like even though the API will take in any reals, these logically should only be positive floats. Adding 
```min="0" onkeypress="return ((event.charCode >= 48 && event.charCode <= 57) || event.charCode === 46)"```
should do the trick here.

**Review by ericcccsliu — CHANGES_REQUESTED** [review]
Amazing work! Really love how you incorporated both the light and dark mode designs, though just a heads up that we might change the designs to prefer one or the other in the future. Just a few tiny details that I think should be changed, and then we'll be all set to merge this in.

File: pages/form.tsx (hunk 8 above)
Comment (AndrewLester) [reply pages/form.tsx]: This is a good catch, we could even restrict the angle input with a `max=359`. I think the `type=number` attribute should be enough instead of the `onkeypress`.

Comment (ericcccsliu) [reply pages/form.tsx]: I think the `onkeypress` is necessary here unfortunately-with just `min=0` you're unable to use the arrows in the input to go below zero, but negatives can be typed through keyboard input just fine. Adding the `onkeypress` basically forbids the '-' sign from being entered.

If there's a more elegant way to do this that'd be great, but this is the only way I've been able to get our expected behavior for now.

Comment (AndrewLester) [reply pages/form.tsx]: That's true... on mobile chrome for me at least the virtual keyboard doesn't even show a `-` for some reason. Also, if we apply a red border color for "[invalid](https://tailwindcss.com/docs/hover-focus-and-other-states#form-states)" (value < min or > max) it could indicate the issue (or a max issue). Also, `min` will prevent submission if the value they input is lower than it. We can use `invalid` in combination with [placeholder-shown](https://tailwindcss.com/docs/hover-focus-and-other-states#placeholder-shown) to show the red border only if they have something input (rather than a currently blank required field).

That all might be overkill though. I'm fine with the keypress solution if we can turn those keycode numbers into constants or replace with them with their string representation.

Comment (ashayp22) [reply pages/form.tsx]: `onKeyPress` is now [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event). Nevertheless, there are other ways to restrict input to positive floats between 0 and 359 (inclusive).

**Review by AndrewLester — APPROVED** [review]

File: pages/form.tsx [hunk 9]
```diff
@@ -0,0 +1,109 @@
...
+      <div className="my-8">
+        <h2 className={h2ClassName}>Solar panel angle tilt</h2>
+        <p className={pClassName}>(Degrees above the horizontal)</p>
+        <input
+          required
+          name="solarPanelAngleTilt"
+          type="number"
+          placeholder="30º"
+          min="0"
+          max="359"
```
ericcccsliu: would change this max to 360 just to encapsulate all floats between 359 and 360

ericcccsliu (approved):
Looks good!

PR #37, opened 2023-02-24
Create Graph Component

### LINKED ISSUE #18: Create graph component
## Description
Create a preliminary graph component to display forecast

Screen Shot 2023-02-21 at 12 25 29 PM

## Todos

- [x] acquire [forecast data](https://github.com/openclimatefix/pv-sites-mobile/blob/main/data/pv-forecast.json)
- [x] display forecast data using [Recharts](https://recharts.org/en-US/examples)
- [x] style according to designs in [Figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=188%3A1116&t=TONZu7WrzKNVaQsB-1)

## Useful links

- [Figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=188%3A1116&t=TONZu7WrzKNVaQsB-1)
- [Recharts](https://recharts.org/en-US/examples)
- [forecast data](https://github.com/openclimatefix/pv-sites-mobile/blob/main/data/pv-forecast.json)

## Time estimate
monday 2/27

rohanvan123:
## Description
For this PR I implemented the `<Graph/>` component, which currently displays a ReCharts `ResponsiveContatiner` component. The current design utilizes the fake forecast data we set up initially. It first makes a mock api call to retrieve the forecast data, then parses the data into a list of `ForecastDataPoint` objects. The ReCharts graph displays time on the x-axis and generated KW on the x-axis. Overall the most challenging part was firstly parsing the data so that it could be graphed and working with ReCharts styling to achieve a look as close as possible to the Figma design.

Fixes #18 

## Todos
- Code Quality Check
- Ensuring Styling is up to standards (may need minor adjustments)
- Ensuring that the component will work with real data garnered from OCF's API.

## Screenshots
Screenshot 2023-02-28 at 3 33 21 PM


components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
...
+  LineChart,
+  ResponsiveContainer,
+  Tooltip,
+  XAxis,
+  YAxis,
+} from 'recharts';
+import { LegendLineGraphIcon } from '@openclimatefix/nowcasting-ui.icons.icons';
+import useSWR, { Fetcher } from 'swr';
+
+interface forecastDataPointProps {
```
AndrewLester: Convention is to capitalize all interface definitions. Do this for the others as well.

```suggestion
interface ForecastDataPointProps {
```

components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
...
+  );
+  tempData.forecast_values.map(({ target_datetime_utc }) => {
+    target_datetime_utc = Date.parse(target_datetime_utc as string);
+  });
+  return tempData as forecastDataProps;
+};
+
+const Graph = () => {
+  const { data, error, isLoading } = useSWR(
+    'http://localhost:3000/api/sites/pv_forecast/b97f68cd-50e0-49bb-a850-108d4a9f7b7e',
```
AndrewLester: Since we'll ultimately want this line to fetch data from their public API, maybe create an environment variable called `NEXT_PUBLIC_API_BASE_URL` and set it to `http://localhost:3000` locally. Then, in production, we can change the variable value to the real base url.

Also, the site UUID at the end of this URL will need to be dynamic since the graph should work for any site. Perhaps pass the site uuid in as a prop.

```suggestion
    `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/sites/pv_forecast/${siteUUID}`,
```

components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
...
+    fetcher
+  );
+
+  const formatter = new Intl.DateTimeFormat(['en-US', 'en-GB'], {
+    hour: 'numeric',
+    minute: 'numeric',
+  });
```
AndrewLester: You can create this outside the component since it doesn't use any component internals. Then it won't be recreated on every render.

components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const truncateDecimals = (number: number, digits: number) => {
+    var multiplier = Math.pow(10, digits),
+      adjustedNum = number * multiplier,
+      truncatedNum = Math[adjustedNum < 0 ? 'ceil' : 'floor'](adjustedNum);
+
+    return truncatedNum / multiplier;
+  };
```
AndrewLester: This might not be necessary with use of [toFixed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed)

components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const MAX = data
+    ? Math.max.apply(
+        Math,
+        data.forecast_values.map(
+          (obj: forecastDataPointProps) => obj.expected_generation_kw
+        )
+      )
```
AndrewLester: `Math.max.apply(Math, array)` is equivalent to `Math.max(...array)`. The `...` will dump the array contents into positional function arguments.
```suggestion
    ? Math.max(
        ...data.forecast_values.map(
          (obj: forecastDataPointProps) => obj.expected_generation_kw
        )
      )
```

components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
...
+    : 0;
+
+  const tickArray = [0, MAX / 4, MAX / 2, (3 * MAX) / 4, MAX];
+
+  return (
+    <>
+      <div className=" w-screen h-[300px] bg-[#444444] rounded-lg">
+        <div className="flex ml-[5%] mt-[20px] mb-[20px] text-sm">
+          <LegendLineGraphIcon className="text-[#FFD053]"></LegendLineGraphIcon>
```
AndrewLester: Can this be self closing?
```suggestion
          <LegendLineGraphIcon className="text-[#FFD053]" />
```

components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
...
+        data.forecast_values.map(
+          (obj: forecastDataPointProps) => obj.expected_generation_kw
+        )
+      )
+    : 0;
+
+  const tickArray = [0, MAX / 4, MAX / 2, (3 * MAX) / 4, MAX];
+
+  return (
+    <>
```
AndrewLester: Since this component's root element is a fragment, and the fragment only has one child, the fragment can be removed.

pages/_app.tsx
```diff
@@ -20,6 +21,7 @@ export default function App({ Component, pageProps }: AppProps) {
         className={`${inter.className} bg-white dark:bg-black flex flex-col items-center justify-start px-10 min-h-screen"`}
       >
         <Component {...pageProps} />
+        <Graph></Graph>
```
AndrewLester: Can this be self closing?
```suggestion
        <Graph />
```

pages/_document.tsx
```diff
@@ -11,3 +11,5 @@ export default function Document() {
     </Html>
   );
 }
+
+// initial
```
AndrewLester:
```suggestion

```

components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
+import React, { useEffect } from 'react';
```
AndrewLester: `import React from 'react';` is not required with Next.js
```suggestion
import { useEffect } from 'react';
```

AndrewLester (changes requested):
Really great work on this, it looks super clean in dev. Especially the legend 👀

components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
...
+const fetcher: Fetcher<forecastDataProps> = async (url: string) => {
+  const tempData: unparsedForecastDataProps = await fetch(url).then((res) =>
+    res.json()
+  );
+  tempData.forecast_creation_datetime = Date.parse(
+    tempData.forecast_creation_datetime as string
+  );
+  tempData.forecast_values.map(({ target_datetime_utc }) => {
+    target_datetime_utc = Date.parse(target_datetime_utc as string);
+  });
```
ericcccsliu: having `as string` here isn't really ideal—we shouldn't really be using `as` without some sort of validation that our element actually _is_ a string. using an `if` statement for type validation here would fix this—by narrowing down types, we can avoid using `as`.  
```suggestion
  if (typeof tempData.forecast_creation_datetime === "string") {
        tempData.forecast_creation_datetime = Date.parse(tempData.forecast_creation_datetime);
    } else {
        throw new Error("Data contains values with incompatible type")
    }
  tempData.forecast_values.map(({ target_datetime_utc }) => {
    if (typeof target_datetime_utc === "string") {
        target_datetime_utc = Date.parse(target_datetime_utc);
    } else {
        throw new Error("Data contains values with incompatible type")
    }
  });
```

pages/_app.tsx
ericcccsliu: we should probably actually remove this from `_app.tsx`—having the component here is good for testing purposes, but we don't want to merge it into main like this!

components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
...
+          (obj: forecastDataPointProps) => obj.expected_generation_kw
+        )
+      )
+    : 0;
+
+  const tickArray = [0, MAX / 4, MAX / 2, (3 * MAX) / 4, MAX];
+
+  return (
+    <>
+      <div className=" w-screen h-[300px] bg-[#444444] rounded-lg">
```
ericcccsliu: change to `w-full` rather than `w-screen` so that when this is implemented, it will only take the width of its container!

components/Graph.tsx
```diff
@@ -0,0 +1,150 @@
...
+        </div>
+
+        <ResponsiveContainer className="mt-[50px]" width="100%" height={200}>
+          <LineChart
+            data={data?.forecast_values}
+            margin={{
+              top: 0,
+              right: 10,
+              left: -25,
+              bottom: 50,
```
ericcccsliu: I think we should decrease the bottom margin to make this component look more similar to our designs. `bottom: 20` looks reasonable on my end.

ericcccsliu (changes requested):

ericcccsliu (commented):
to wrap up this PR, it'd be awesome if you could do a quick `git pull origin main` and slot this graph component into our `dashboard` page! that way, we can make sure the styles are perfect for the dashboard, and we won't have to open up a new issue later on for integration of this component. 

also, as a general reminder, make sure to start writing your git commit messages in the **imperative mood**—i.e., "finish basic styling for graph component" rather than "finished basic styling for graph component." it's a really tiny change, but it's great to get in the habit of following convention!

components/Graph.tsx
ericcccsliu: also, we can avoid using this one-off styling in `bg-[#444444]`—if you do a pull from main, this color is actually in our tailwind config as `ocf-gray-1000`

package.json
```diff
@@ -12,10 +12,12 @@
     "style": "yarn format && yarn lint"
   },
   "dependencies": {
-    "@auth0/nextjs-auth0": "^1.8.0",
+    "@auth0/nextjs-auth0": "^2.2.1",
```
AndrewLester: While our code will work with this version of next auth, certain bit.dev components OCF has provided required the version we currently have in the repo. For now, I think we should keep it downgraded in case we need to use these, and consider upgrading later.

Make sure to run yarn install after changing this back.

pages/_app.tsx
```diff
@@ -1,6 +1,6 @@
 import '~/styles/globals.css';
 import type { AppProps } from 'next/app';
-import { UserProvider } from '@auth0/nextjs-auth0';
+import { UserProvider } from '@auth0/nextjs-auth0/client';
```
AndrewLester: This is required after upgrading next auth, but we want to keep the current repo version of next auth, so no need to do this.

AndrewLester (approved):
Just two quick things, then you can merge pending Eric's review.

ericcccsliu (approved):
Looks good, great work making the required changes and integrating the graph into our dashboard page!

PR #49, opened 2023-03-01
Fix links and re-add nav and header

AndrewLester:
## Status:

:rocket: Ready

## Description

Apparently if a `<Link` tag has children that aren't just text, it needs to have an `<a` as a child to actually render as a link appropriately. I added this to the side bar links, and re-added the side bar. I also fixed some attribute warnings.


ericcccsliu (changes requested):
Looks good—only thing is that right now the placement of the Layout component in _app.tsx causes the navbar to show up in all pages, even pages where we don't want the navbar to be present. We can fix this by adding a getLayout function to the pages where we want the layout to be applied (just the dashboard page for now) and utilizing our current Layout component there.

ericcccsliu (approved):
Approved—will address page-specificity with sidebar in issue #51

PR #56, opened 2023-03-02
Create Near Future Threshold Graph

### LINKED ISSUE #48: Near future threshold based graph
## Description

Create the near future threshold based graph:
image

## Todos

- [x] Pull from [this mock api endpoint](https://github.com/openclimatefix/pv-sites-mobile/blob/43111b8af516c177e314e9cbe659f2e8dea4bd09/pages/api/%5B...mockApiRoute%5D.ts#L40) using [swr](https://swr.vercel.app/)
- [ ] Create the dashed graph line for the actual data using a [tiny area chart](https://recharts.org/en-US/examples/TinyAreaChart). You can get a gradient effect going from [this example](https://recharts.org/en-US/api) (though you want a reverse of theirs)
- [ ] Add a threshold line to the chart at 0.7 for now.
- [ ] Create a custom chart legend like in the figma. You can use an svg or just divs for the example lines
- [ ] Determine whether solar activity is increasing, and for how long ([RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat)). Make sure the intervals are in sorted order and assume the first data point (chronologically) is now (though it may not be in the future)
- [ ] Create a label for the threshold line (2000 kW in the figma). Write 0.7 for now, since our data is only 0-1.
- [ ] Use the arrow icon from [heroicons](https://heroicons.com/) (copy JSX)
- [ ] Apply the circle with ring at the "now" point in the graph (the start, for now). This may be more difficult to do... consult links

## Useful links

- [Figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=188%3A1116&t=fRhtAx2p3F4VCA0m-0)
- [Recharts api docs](https://recharts.org/en-US/api)
- [Existing nowcasting app's chart code](https://github.com/openclimatefix/nowcasting/tree/main/apps/nowcasting-app/components/charts). Look at this code for examples on how to render very intricate charts

## Time estimate

We'll try before MPC.

ashayp22:
## Description

Create the near future threshold based graph:

Fixes #48 

## Todos

- [x] Pull from [this mock api endpoint](https://github.com/openclimatefix/pv-sites-mobile/blob/43111b8af516c177e314e9cbe659f2e8dea4bd09/pages/api/%5B...mockApiRoute%5D.ts#L40) using [swr](https://swr.vercel.app/)
- [x] Create the dashed graph line for the actual data using a [tiny area chart](https://recharts.org/en-US/examples/TinyAreaChart). You can get a gradient effect going from [this example](https://recharts.org/en-US/api) (though you want a reverse of theirs)
- [x] Add a threshold line to the chart at 0.7 for now.
- [x] Create a custom chart legend like in the figma. You can use an svg or just divs for the example lines
- [x] Determine whether solar activity is increasing, and for how long ([RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat)). Make sure the intervals are in sorted order and assume the first data point (chronologically) is now (though it may not be in the future)
- [x] Create a label for the threshold line (2000 kW in the figma). Write 0.7 for now, since our data is only 0-1.
- [x] Use the arrow icon from [heroicons](https://heroicons.com/) (copy JSX)
- [x] Apply the circle with ring at the "now" point in the graph (the start, for now). This may be more difficult to do... consult links

## Screenshots

WIP


components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
...
+  ReferenceLine,
+} from 'recharts';
+import useSWR, { Fetcher } from 'swr';
+import {
+  FutureThresholdLegendIcon,
+  UpArrowIcon,
+  DownArrowIcon,
+  LineCircle,
+} from './icons/future_threshold';
+interface ForecastDataPointProps {
```
AndrewLester: New line above this interface just to separate it from imports.

In terms of the names of these interfaces, since they aren't being used to type the props of a React component, and instead they're core data types our app is using from fetching and processing, I think we can remove the "Props" suffix from this and the ones below.

```suggestion

interface ForecastDataPoint {
```

components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+  tempData.forecast_values.map(({ target_datetime_utc }) => {
+    if (typeof target_datetime_utc === 'string') {
+      target_datetime_utc = Date.parse(target_datetime_utc);
+    } else {
+      throw new Error('Data contains values with incompatible types');
+    }
+  });
+  return tempData as ForecastDataProps;
+};
```
AndrewLester: This file is pretty large, and that makes sense since it's supposed to do a lot. Because it's so big though, I think separating the data fetching/parsing part of this code into a separate file could improve the readability of this component and make it easier to tweak down the road if necessary. A start would be to move this `fetcher` function and the corresponding `useSWR` hook in the component to another file near by, renaming the hook to `useFutureGraphData`. This would add a level of indirection to the data for this component, but I think it could be worth it by reducing file size.

components/icons/future_threshold.tsx
```diff
@@ -0,0 +1,74 @@
...
+      fill="#E0E0E0"
+    />
+    <path
+      d="M73 5C67.9282 5 63.8011 5 59.7238 5C55.6464 5 56.0689 5 51.4199 5C50.2247 5 46 5 46 5"
+      stroke="#E0E0E0"
+      stroke-width="2"
+      stroke-dasharray="2 2"
```
AndrewLester: For some reason React is giving me warnings on the server unless these are changed as so:
```suggestion
      strokeWidth="2"
      strokeDasharray="2 2"
```

components/icons/future_threshold.tsx
```diff
@@ -0,0 +1,74 @@
...
+      fill="#FFD053"
+    />
+    <path
+      d="M73 15C67.9282 15 63.8011 15 59.7238 15C55.6464 15 56.0689 15 51.4199 15C50.2247 15 46 15 46 15"
+      stroke="#FFD053"
+      stroke-width="2"
+      stroke-dasharray="2 2"
```
AndrewLester: Same as above.
```suggestion
      strokeWidth="2"
      strokeDasharray="2 2"
```

components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
...
+  };
+
+  /**
+   * Renders two types of labels for the graph:
+   * - A circle for the current time
+   * - A threshold label
+   * @param props data about the point, such as x and y position on the graph
+   * @returns SVG element
+   */
+  const renderGraphLabels = (props: any) => {
```
AndrewLester: I think this function could be split into 2 separate ones since it has 2 distinct goals. Maybe this current function could just show the line circle label, and the label for the threshold could be put into another function, which is then passed to another `<LabelList>` component. If that works, that is, I haven't tested.

components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
...
+        return (
+          <g>
+            <LineCircle x={x} y={y} />
+          </g>
+        );
+      } else if (index === 0) {
+        return (
+          <svg
+            x={x - 30}
+            y={-78.95 * graphThreshold + 76.84}
```
AndrewLester: While these magic numbers are literally magic, can they be turned into constants that say a little about what they do?

components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
...
+              fill="#FFD053"
+            />
+          </svg>
```
AndrewLester: Since we're already in an `<svg>` context and since this `<path>` is just for text, can we switch this to something like (these styles aren't exact but you get the gist):
```suggestion
          <text
            fill="#FFD053"
            x={x - 30}
            y={-78.95 * graphThreshold + 76.84}
            style={{ position: 'relative', marginLeft: '30px' }}
          >
            2000 kW
          </text>
```

components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
...
+   * @returns SVG gradient
+   */
+  const generateGraphGradient = () => {
+    if (!isLoading && data) {
+      const aboveThreshold = data.forecast_values.some(
+        (forecast) => forecast.expected_generation_kw > graphThreshold
+      );
+
+      if (aboveThreshold) {
+        let gradientPercentage = -107.5 * graphThreshold + 99.25;
```
AndrewLester: Perhaps separate magic numbers into variables.... but this function is already pretty clean and again there's not much that can make these less magic, they're just styles I guess.

components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
...
+   * Renders two types of labels for the graph:
+   * - A circle for the current time
+   * - A threshold label
+   * @param props data about the point, such as x and y position on the graph
+   * @returns SVG element
+   */
+  const renderGraphLabels = (props: any) => {
+    const { x, y, index } = props;
+
+    if (!isLoading && data && data.forecast_values.length > 0) {
```
AndrewLester: There's often a check for `!isLoading && data` at the top of these separate rendering functions. What if you instead check for `isLoading || !data` and return early to save an indent level in all cases?

components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
...
+                dataKey="expected_generation_kw"
+                content={renderGraphLabels}
+              />
+            </Area>
+            <ReferenceLine
+              y={graphThreshold}
+              strokeWidth={2}
+              stroke="#FFD053"
+              strokeDasharray="2"
+            ></ReferenceLine>
```
AndrewLester: Can this self close?
```suggestion
            />
```

AndrewLester (changes requested):
Simply put, this looks amazing. The calculations in here are very impressive also and look like they handle all cases, so nice work. A few small changes + one larger thought.

components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
+import {
```
ericcccsliu: Let's rename this component—technically, the other `Graph` component we have is also a future graph (since it deals with the future forecast). Something like `ThresholdGraph` should suffice!

components/FutureGraph.tsx
ericcccsliu: most of this is actually copied over from the `Graph` component—maybe we could target the shared bits to abstract

pages/dashboard.tsx
```diff
@@ -10,6 +11,9 @@ const Dashboard = () => {
       <div className="flex flex-row w-full justify-start">
         <Warnings />
       </div>
```
ericcccsliu: Let's get rid of this dashboard module since it's just a stand-in—there are modules in our designs that should go here but putting your module at the top should be good for now

components/FutureGraph.tsx
ericcccsliu: just to keep this file size somewhat reasonable, can we abstract this out?

components/FutureGraph.tsx
ericcccsliu: I don't like the `props: any` here but I spent the last half hour reading recharts' source code and I'm pretty sure `any` here is the best we can do 😭

components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
...
+          ) {
+            return solarDecreasingText(formattedDate);
+          }
+        }
+        currIndex += 1;
+      }
+    }
+
+    return '';
+  };
```
ericcccsliu: Consider abstracting this out into utils or another suitable file—this file getting pretty long, and we might want to use this function in other future modules for our dashboard!

components/FutureGraph.tsx
```diff
@@ -0,0 +1,318 @@
...
+const FutureGraph = () => {
+  const { data, isLoading } = useSWR(
+    `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/sites/pv_forecast/${siteUUID}`,
+    fetcher
+  );
+
+  /* The index of the current time in the forecast data */
+  const getCurrentTimeForecastIndex = () => {
+    return 9;
+  };
```
ericcccsliu: Could we actually implement this as a function that gets the closest index to our current time in the forecast data?

ericcccsliu: It might not look great because a lot of the time the dot will either be at the beginning or the end due to the nature of our fake data, but this will be very useful once we hook this up with our API

ericcccsliu (changes requested):
Looks great—just a few code quality suggestions in the comments, and one bigger ask.

AndrewLester (approved):

ericcccsliu (approved):

PR #57, opened 2023-03-02
Implement form context provider and post data

### LINKED ISSUE #45: Form context and submission to mock endpoint
## Description

Add a context provider and hook system to store form data for both form pages, and send all the aggregated data to the right endpoint to save the data for a user. After this data is saved, redirect them to the dashboard (for now).

## Todos

- [x] Create a context provider/hook system to store and update form data for the location input and panel details pages
- [x] Upon submission on the panel details page, send this data to the api endpoint
- [x] After success, redirect them to the right page

## Useful links

- https://reactjs.org/docs/context.html
- Amit's context presentation

anyaparekh:
## Status: 
:rocket: Ready

## Description
* Creates custom form context and provider.
* This will allow data from both the site location page and site details page to be aggregated and posted at once when the next button on the site details page is clicked.
* Added form context to map and form pages.
* Added redirects between map, form, and dashboard.

Fixes #45 

## Todos
* Handle errors from post requests.

## Screenshots
image


lib/context/form_context.tsx
```diff
@@ -0,0 +1,88 @@
...
+    headers: {
+      'content-type': 'application/json',
+    },
+  }).then((res) => res.json());
+}
+
+const FormContext = React.createContext<Form | null>(null);
+
+const FormProvider: FC<FormProviderProps> = ({ children }) => {
+  const { trigger } = useSWRMutation('/api/[...mockApiRoute]', sendRequest);
```
ericcccsliu: we actually want to be sending POST requests to the `/api/sites` route. also, I think we should be including the entire url here, as our API base URL will be different once we move away from mockdata. if you look in the `Graph` component there should be an example of how to use our current environment variables to specify the correct URL.

lib/context/form_context.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const data: FormPostData = {
+      site_uuid: 1,
+      client_name: 'name',
+      client_site_id: 1,
+      client_site_name: 'site_name',
+      latitude: latLong[0],
+      longitude: latLong[1],
+      installed_capacity_kw: capacity,
+      created_utc: 'utc_create',
+      updated_utc: 'utc_update',
```
ericcccsliu: these can be set right now—both can probably be set to the current time in UTC. in the future, we'll want to differentiate between creating a site and updating a site, and created_utc might well differ from updated_utc as a result, but currently, setting both to be the same thing should be fine. we'll address that once multi-site functionality is added and we gain easier access to the real API.

lib/context/form_context.tsx
ericcccsliu: ok also apparently my mock api route does not work properly—there should be a fix available when #62 gets merged in.

lib/context/form_context.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const setFormData = (direction: number, tilt: number, capacity: number) => {
+    const data: FormPostData = {
+      site_uuid: 1,
+      client_name: 'name',
+      client_site_id: 1,
+      client_site_name: 'site_name',
+      latitude: latLong[0],
+      longitude: latLong[1],
+      installed_capacity_kw: capacity,
```
ericcccsliu: capacity rn might be NaN—which is falsy—as it is not required in our form. while this is the behavior we want, the API currently still doesn't allow for null capacity values. to get around this for now, lets put in an arbitrary constant "sentinel" value when the user doesn't have capacity specified.

lib/context/form_context.tsx
```diff
@@ -0,0 +1,88 @@
...
+      value={{
+        latLong,
+        setFormData,
+        setMapData,
+      }}
+    >
+      {children}
+    </FormContext.Provider>
+  );
+};
```
ericcccsliu, line 77: great job using context wooooo  👍 🥇 💯

lib/context/sidebar_context.tsx
```diff
@@ -23,14 +23,16 @@ const AppProvider: FC<AppProviderProps> = ({ children }) => {
     setIsSidebarOpen(false);
   };
   return (
-    <AppContext.Provider value={{ isSidebarOpen, openSidebar, closeSidebar }}>
+    <SidebarContext.Provider
```
ericcccsliu, line 26: thanks for renaming these!!!

ericcccsliu (changes requested):

ericcccsliu (commented):
Another comment—could we make sure that the relevant links in the sidebar link to the location input page rather than the form input page?

lib/context/form_context.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+interface Form {
+  latLong: [number, number];
+  setFormData: (direction: number, tilt: number, capacity: number) => void;
+  setMapData: (lat: number, long: number) => void;
+}
+
+interface FormProviderProps {
+  children: ReactNode | ReactNode[];
+}
```
AndrewLester: This can be deleted, you'll see why lower down.
```suggestion

```

lib/context/form_context.tsx
```diff
@@ -0,0 +1,88 @@
...
+    body: JSON.stringify(arg),
+    headers: {
+      'content-type': 'application/json',
+    },
+  }).then((res) => res.json());
+}
+
+const FormContext = React.createContext<Form | null>(null);
+
+const FormProvider: FC<FormProviderProps> = ({ children }) => {
```
AndrewLester: Instead of manually adding a type for the `children` prop, React provides one you can use in `PropsWithChildren`. You'll need to import this though.
```suggestion
const FormProvider: FC<PropsWithChildren> = ({ children }) => {
```

AndrewLester (changes requested):
Only got a few small comments, this looks really nice. Once Eric's stuff is resolved this should be good to go.

components/NavBar.tsx
```diff
@@ -1,5 +1,5 @@
 import { FC } from 'react';
-import { useSidebarContext } from './context';
+import { useSidebarContext } from '../lib/context/sidebar_context';
```
neha-vard: Should this import path have the same format as the import line in SideBar?

neha-vard (commented):
This looks really good!

components/NavBar.tsx
anyaparekh: yes you're right!

lib/context/form_context.tsx
```diff
@@ -0,0 +1,89 @@
...
+      'content-type': 'application/json',
+    },
+  }).then((res) => res.json());
+}
+
+const FormContext = React.createContext<Form | null>(null);
+
+const FormProvider: FC<PropsWithChildren> = ({ children }) => {
+  const { trigger } = useSWRMutation(
+    `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/sites/`,
```
AndrewLester: These are equivalent, just preference.

```suggestion
    `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/sites`,
```

pages/form.tsx
```diff
@@ -22,16 +24,23 @@ const preventMinus = (e: React.KeyboardEvent<HTMLInputElement>) => {
...
+  const onSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
 
     // TODO: Add schema validation with zod
-    const siteDetailsData = Object.entries(
-      formData
-    ) as unknown as PanelFormDataBody;
+    // const siteDetailsData = Object.entries(
+    //   formData
+    // ) as unknown as PanelFormDataBody;
```
AndrewLester: Can be removed if we're not going with the form data system.
```suggestion

```

lib/context/form_context.tsx
```diff
@@ -0,0 +1,89 @@
...
+      client_site_name: 'site_name',
+      latitude: latLong[0],
+      longitude: latLong[1],
+      installed_capacity_kw: !!capacity ? capacity : sentinel,
+      created_utc: date,
+      updated_utc: date,
+      orientation: direction,
+      tilt: tilt,
+    };
+    trigger(data);
```
AndrewLester, line 60: For later: Putting the `trigger(...)` call into `setFormData` isn't invalid in any way, that is, it should work. My concern is that the function name `setFormData` implies setting some state inside the context, and maybe not actually sending the data to the API. Let me know if you agree, and we'll create a new issue to think about this later.

AndrewLester (approved):
A few small things and one consideration for later.

ericcccsliu (approved):

PR #61, opened 2023-03-03
Add output and yield modules

### LINKED ISSUE #44: Current output and current percent yield modules
## Description
Create current output and current percent yield modules. For output, look at the forecast that is closest to the current time. For percent yield, use output data as well as `installed_capacity_kw` information from the `sites/site_list` endpoint. 

Screen Shot 2023-02-27 at 7 16 30 PM

## Todos

- [x] acquire data within each module component using SWR 
- [x] get the closest forecasted `expected_generation_kw` datapoint 
- [x] calculate percent yield 
- [ ] style according to [figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=389%3A963&t=yMNRyKNs88okH6Dj-1)

## Useful links

- [figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=389%3A963&t=yMNRyKNs88okH6Dj-1)

next devnight 3/2

AndrewLester:
## Description
Create current output and current percent yield modules. For output, look at the forecast that is closest to the current time. For percent yield, use output data as well as installed_capacity_kw information from the sites/site_list endpoint.

Fixes #44

## Screenshots
image


ericcccsliu (changes requested):
looks great! let's refactor those modules to handle data parsing on their own before merging!

anyaparekh (approved):
FIRE 🔥 🔥 🔥

components/CurrentCapacity.tsx
```diff
@@ -0,0 +1,56 @@
...
+  );
+  let installed_capacity_kw = site_list?.site_list[0].installed_capacity_kw;
+
+  return (
+    <NumberDisplay
+      title="Current Capacity"
+      value={`${
+        cur_output != null && installed_capacity_kw != null
+          ? (cur_output / installed_capacity_kw)?.toFixed(0)
+          : 'Loading'
```
ericcccsliu: a less verbose way to do this would be using the '!!' operator, which casts 'truthy' or 'falsy' values to true and false, respectively 

also, instead of 'Loading' maybe put N/A, since a null value doesn't necessarily just correspond to loading here. we'll also think more about what should go here in that case, but this should work for now. 
```suggestion
        !!cur_output && !!installed_capacity_kw
          ? (cur_output / installed_capacity_kw)?.toFixed(0)
          : 'N/A'
```

components/CurrentOutput.tsx
```diff
@@ -0,0 +1,47 @@
...
+    time = new Date(actual_outputs[i].datetime_utc).getTime();
+    if (Math.abs(cur_time - time) < min) {
+      min = cur_time - time;
+      index = i;
+    }
+  }
+  return actual_outputs[index].actual_generation_kw;
+}
+
+const siteUUID = 'b97f68cd-50e0-49bb-a850-108d4a9f7b7e';
```
ericcccsliu, line 25: we should consider putting this into an ENV variable for testing but maybe that'd be overkill

components/CurrentOutput.tsx
```diff
@@ -0,0 +1,47 @@
...
+  console.log(pv_actual);
+
+  let cur_output = pv_actual
+    ? GetCurrentOutput(pv_actual?.pv_actual_values)
+    : null;
+
+  return (
+    <NumberDisplay
+      title="Current Output"
+      value={`${cur_output != null ? cur_output.toFixed(0) : 'Loading'} kW`}
```
ericcccsliu: change loading here too for now

ericcccsliu: actually let's make the value here say 'loading' if isLoading on the useSWR call is true, and then say 'N/A' if everything is loaded but the value is still falsy

components/CurrentCapacity.tsx
ericcccsliu: Actually, lets wrap this all in an isLoading ternary, so if the data is actually loading, we display 'Loading...'

ericcccsliu (commented):
Looks great! Just a few tiny changes, and we can merge this in

ericcccsliu (approved):

PR #64, opened 2023-03-03
Use custom auth required wrapper on dashboard page

### LINKED ISSUE #58: Keyboard navigation exposes sidebar when hidden
**Describe the bug**
Using keyboard navigation, users can tab into the sidebar and enter the links within the sidebar, all while the sidebar remains closed
**To Reproduce**
1. go to any page where the navbar/sidebar exist
2. focus on the sidebar open/close icon 
3. hit tab to focus on following elements 

**Expected behavior**
hitting tab when focused on the sidebar icon, while the sidebar is closed, should focus the user on the next visible element on the page

**Additional context**

AndrewLester:
## Status:

:rocket: Ready

## Description

Use our custom auth required wrapper function on the dashboard page to allow the dashboard page to work in preview deployments.

Also, reorganize form pages and add a redirect from '/' to either '/dashboard' or '/form/location' if the user has no sites. This needs to wait until #57 is in. The redirect will not work in preview.

Also fixes #58, and I've determined the flickering hamburger button is not going to be worth it to fix... though considering the bottom nav will have the same flickering problem, we may want to default to showing the icons and fade them out instead, since most users will have a site.


components/NavBar.tsx
```diff
@@ -1,42 +1,17 @@
...
+import { Site } from '~/lib/types';
+import { MenuLogo, NowcastingLogo } from './icons/navbar_icons';
 
 const NavBar: FC = () => {
   const { isSidebarOpen, openSidebar } = useSidebarContext();
-  const { data, isLoading } = useSWR(
-    `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/sites/site_list`,
-    fetcher
+  const { data, isLoading } = useSWR<{ site_list: Site[] }>(
+    `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/sites/site_list`
```
ericcccsliu, line 10: great use of the custom default fetcher here to simplify our code!

lib/swr.ts
```diff
@@ -0,0 +1,4 @@
+export async function fetcher(url: string) {
+  const res = await fetch(url);
+  return await res.json();
+}
```
ericcccsliu, line 4: oh this is sick

pages/index.tsx
```diff
@@ -22,4 +24,32 @@ export default function Home(this: any) {
...
+      {
+        headers: {
+          Authorization: `Bearer ${session?.accessToken}`,
+        },
+      }
+    ).then((res) => res.json())) as { site_list: Site[] };
+
+    const destination =
+      site_list.length === 0 ? '/form/location' : '/dashboard';
+
```
ericcccsliu, line 46: this is great

ericcccsliu (approved):
Overall: looks good to me. Tested out all the routes, form functionality (including POST), redirect from index w/no sites, and it all worked perfectly.

PR #78, opened 2023-03-06
Fix threshold calculation for graph gradient

ashayp22:
## Description

Updates the calculation for the gradient of the threshold graph by taking into account the proportion between the maximum solar output and the solar threshold.

### DIFF
--- a/components/graphs/ThresholdGraph.tsx
+++ b/components/graphs/ThresholdGraph.tsx
@@ -23,8 +23,8 @@ import {
   Value,
 } from 'lib/utils';
 
-/* Represents the threshold (2000 kWh) for the graph */
-const graphThreshold = 0.4;
+/* Represents the threshold for the graph */
+const graphThreshold = 0.7;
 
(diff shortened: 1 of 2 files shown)


AndrewLester (approved):
I think Eric knows better what this does.

components/graphs/ThresholdGraph.tsx
```diff
@@ -133,7 +134,9 @@ const ThresholdGraph = () => {
             ({ expected_generation_kw }) => expected_generation_kw
           )
         );
-        let gradientPercentage = -107.5 * graphThreshold + 96;
+
+        let gradientPercentage =
+          100 - (graphThreshold / maxExpectedGenerationKW) * 100;
```
ericcccsliu, line 139: this does seem to make more sense

ericcccsliu (approved):
looks good-tested a ton of different threshold values and the gradient worked for every one! great work!!!

PR #79, opened 2023-03-06
Fix memory leak in location input page

### LINKED ISSUE #68: Memory leak in form submission sequence
**Describe the bug**
On form submission, we are occasionally getting this error
image

**To Reproduce**

1. Go to the location input page
2. input your location
3. input your panel details
4. see error in console (occasionally) 

**Expected behavior**
no memory leak

ericcccsliu:
## Status:
:rocket: Ready

## Description

Fixes the memory leak in the location input page by disabling the next button while any movement in the map is occurring. Also adds a fade to the button to make it look nice 💯 

Fixes #68


components/LocationInput.tsx
```diff
@@ -82,6 +82,27 @@ const MapBoxInput: FC<PropsWithChildren<MapBoxInputProps>> = ({
...
+          setZoom(map.current!.getZoom());
+          setIsSubmissionEnabled(map.current!.getZoom() > zoomLevelThreshold);
+          updateMarker(
+            marker,
+            map.current!,
+            zoomLevelThreshold,
+            newLng,
+            newLat
+          );
+        });
```
AndrewLester: If you move the arrow function passed into the `off` call (the lines this comment refers to) to a separate declaration in the useEffect and name it `handler` or something, you can then use the same declaration when `on` is called a few lines above. This will effectively remove the handler when the component unmounts, and might fix the memory leak issue? It also reduces the code duplication.

ericcccsliu: oops i hadn't pushed yet this is very old

ericcccsliu: also doing that actually funnily just breaks the map—the marker just fails to show up

AndrewLester: Dang alright

AndrewLester (approved):

PR #81, opened 2023-03-07
Integrate graphs with API deployment

### LINKED ISSUE #67: Integrate graphs with API deployment
## Description
We'd like to start integrating our app with the API deployment. Currently, the API returns a forecast for the next three days, which is a much longer time period than we want for our near-term graph. We'll need to write a helper function to return a portion of this data over a shorter time period, and use that function to feed shorter-term data into our graphs. 

For the near-term graph, we can probably just display the forecast for the current day. @AndrewLester and I were thinking that we could end this graph at 8 pm or so and start displaying the next day's forecast after then. For the longer-term graph further down in the dashboard, we could maybe display the forecast for the next 48 hours, or maybe even just include all the data that the API gives us. 

## Todos

- [x] create separate utils file just for graphs, reorganize current utils file 
- [x] create a helper function to return a subset of API forecast data over a specified time range 
- [x] restrict time ranges displayed in both of our graphs using this helper function 
- [x] reorganize graphs into own directory
- [x] also put an example forecast from the API into our mock data for testing purposes—there might be cases where we still go back to our mock routes, as we might not want to POST to the API for now 

## Useful links

- [api docs](http://pvsite-development-api-sites.eu-west-1.elasticbeanstalk.com/docs)
- [currently working sites](https://www.notion.so/h4iuiuc/pv-sites-api-notes-c3cdacf844d3476c992272cc8b4ef5c2)

## Time estimate
tbd

neha-vard:
## Description

We'd like to start integrating our app with the API deployment. Currently, the API returns a forecast for the next three days, which is a much longer time period than we want for our near-term graph. We'll need to write a helper function to return a portion of this data over a shorter time period, and use that function to feed shorter-term data into our graphs.

For the near-term graph, we can probably just display the forecast for the current day. @AndrewLester and I were thinking that we could end this graph at 8 pm or so and start displaying the next day's forecast after then. For the longer-term graph further down in the dashboard, we could maybe display the forecast for the next 48 hours, or maybe even just include all the data that the API gives us.

Fixes #67


components/graphs/Graph.tsx
```diff
@@ -8,16 +8,25 @@ import {
...
-import { formatter } from 'lib/utils';
+import { formatter, forecastDataOverDateRange } from 'lib/graphs';
 import { useSiteData } from 'lib/hooks';
 import { FC } from 'react';
 
 const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
   const { forecastData } = useSiteData(siteUUID);
 
-  const maxGeneration = forecastData
+  let endDate = new Date();
```
AndrewLester:
```suggestion
  const endDate = new Date();
```

components/graphs/ThresholdGraph.tsx
```diff
@@ -31,14 +32,30 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
   const { forecastData, latitude, longitude, isLoading } =
     useSiteData(siteUUID);
   const { currentTime } = useTime(latitude, longitude);
-
+  let currentDate = new Date();
+  let endDate = new Date();
+  currentDate.setHours(8);
+  currentDate.setMinutes(0);
+  currentDate.setSeconds(0);
+  currentDate.setMilliseconds(0);
```
AndrewLester: Maybe instead of computing these in the components, something like this could work?
```
  // Defined somewhere above the component...
  function getGraphEndDate(date: Date) {
    endDate.setHours(20);
    endDate.setMinutes(0);
    endDate.setSeconds(0);
    endDate.setMilliseconds(0);
  }

  // In the component
  const endDate = getGraphEndData(new Date());
```

AndrewLester (approved):

components/graphs/ThresholdGraph.tsx
```diff
@@ -32,13 +33,39 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
+    20
+  );
+
+  const graphData = forecastData
+    ? forecastDataOverDateRange(
+        JSON.parse(JSON.stringify(forecastData.forecast_values)),
+        startDate,
+        endDate
+      )
+    : [];
```
AndrewLester: Smart to default this to `[]`

AndrewLester (approved):

components/graphs/ThresholdGraph.tsx
```diff
@@ -92,17 +117,15 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
    * @returns SVG gradient
    */
   const generateGraphGradient = () => {
-    if (!isLoading && forecastData) {
-      const aboveThreshold = forecastData.forecast_values.some(
+    if (!isLoading && graphData) {
```
neha-vard:
```suggestion
    if (!isLoading && graphData.length > 0) {
```

components/graphs/ThresholdGraph.tsx
```diff
@@ -133,19 +156,15 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
    * @returns the start and end time label on the graph's x-axis
    */
   const renderStartAndEndTime = () => {
-    if (!isLoading && forecastData) {
-      const numForecastValues = forecastData.forecast_values.length;
+    if (!isLoading && graphData) {
```
neha-vard:
```suggestion
    if (!isLoading && graphData.length > 0) {
```

components/BottomNavBar.tsx
```diff
@@ -27,7 +27,7 @@ const BottomNavBar = () => {
   const { asPath } = useRouter();
 
   const { data, isLoading } = useSWR(
-    `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/sites`
+    `${process.env.NEXT_PUBLIC_API_BASE_URL_GET}/sites`
   );
```
ericcccsliu, line 31: great catch

env.d.ts
```diff
@@ -9,6 +9,7 @@ namespace NodeJS {
     AUTH0_CLIENT_SECRET: string;
     NEXT_PUBLIC_MAPBOX_PUBLIC: string;
     NEXT_PUBLIC_AUTH0_LOGOUT_REDIRECT: string;
-    NEXT_PUBLIC_API_BASE_URL: string;
+    NEXT_PUBLIC_API_BASE_URL_POST: string;
+    NEXT_PUBLIC_API_BASE_URL_GET: string;
```
ericcccsliu, line 13: ah thanks for adding these in—keep forgetting that this file exists

components/graphs/ThresholdGraph.tsx
```diff
@@ -32,13 +33,39 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
+    8
+  );
+  const endDate = new Date(
+    currentDate.getFullYear(),
+    currentDate.getMonth(),
+    currentDate.getHours() > 20
+      ? currentDate.getDate() + 1
+      : currentDate.getDate(),
+    20
+  );
```
ericcccsliu: might be worth it to put these into their own functions

lib/graphs.ts
```diff
@@ -0,0 +1,126 @@
...
+  forecastData: ForecastDataPoint[],
+  start_date: Date,
+  end_date: Date
+) => {
+  const start_index = getClosestForecastIndex(forecastData, start_date);
+  const end_index = getClosestForecastIndex(forecastData, end_date);
+  if (forecastData)
+    forecastData = forecastData.slice(start_index, end_index + 1);
+  return forecastData;
+};
```
ericcccsliu, line 96: very elegant way to do this

ericcccsliu (approved):
As a bonus—since you're already working on this, could you merge in main again (should be very easy, no merge conflicts exist) and apply the same restricted dates from the `ThresholdGraph` to the mini graphs in the site cards?

AndrewLester:
Unfortunately I just added one small merge conflict to this, after you merge main. It's quite small, just use your best judgment.

components/graphs/ThresholdGraph.tsx
```diff
@@ -32,13 +35,38 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
+        forecastData
+          ? forecastDataOverDateRange(
+              forecastData.forecast_values,
+              getGraphStartDate(currentTime),
+              getGraphEndDate(currentTime)
+            )
+          : []
+      );
+    }
+  }, [forecastData, currentTime, graphData]);
```
AndrewLester: Instead of a `useEffect` here and a `useState` above, can `graphData` just be a `const` defined to be equal to
```
forecastData
          ? forecastDataOverDateRange(
              forecastData.forecast_values,
              getGraphStartDate(currentTime),
              getGraphEndDate(currentTime)
            )
          : []
```

or does that not work?

AndrewLester (approved):
One last thing, very nice work. This is a pretty big PR!

components/graphs/ThresholdGraph.tsx
```diff
@@ -32,13 +35,38 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
+          forecastData.forecast_values,
+          getGraphStartDate(currentTime),
+          getGraphEndDate(currentTime)
+        )
+      : []
+  );
+
+  useEffect(() => {
+    if (graphData.length === 0) {
+      console.log('updating');
```
ericcccsliu: remove this console.log

components/graphs/Graph.tsx
```diff
@@ -8,19 +8,26 @@ import {
...
-        )
+  const endDate = new Date();
+  endDate.setHours(endDate.getHours() + 48);
+  const graphData = forecastData
+    ? forecastDataOverDateRange(
+        forecastData.forecast_values,
+        new Date(currentTime),
+        endDate
       )
+    : [];
```
ericcccsliu, line 28: if setting this to [] by default breaks things (as in our bug with `ThresholdGraph`), perhaps we could consider just setting this to be `undefined` and checking for that instead of having to do `graphData.length > 0`?

ericcccsliu (approved):
Absolutely massive PR—great work on all of this. A couple choices that we'll have to discuss, but this will definitely be ready to merge in soon.

components/graphs/Graph.tsx
neha-vard: Just tried this out and the threshold label and current time marker are still not rendering

components/graphs/ThresholdGraph.tsx
neha-vard: So until we did this, the ThresholdGraph labels (current time marker, threshold label) weren't showing up. @ashayp22 helped us figure out that the were likely not rendering because at first, graphData did not contain any data until forecastData was read with useSiteData so there was nothing to label, and then the problem (probably) was is that they didn't update when graphData updated.

AndrewLester: Thanks for the info, yeah this works then

components/graphs/Graph.tsx
ericcccsliu: alright keep it like this then

PR #85, opened 2023-03-07
Add back button to panel details page and hide hamburger on mobile

### LINKED ISSUE #60: Back button on form input pages
## Description
Add a back button to the top of our form pages, so that the user may go back and forth between the form input pages

Feel free to make this a button with a call to `router.push(...)` instead of a `<Link` since, honestly, we don't want people to r-click it and do open in new tab since they'd lose all their previous form state.

Do note in these designs that there will be a top nav on the real site:
image

## Todos

- [x] Add back button with a call to `router.push(...)`

## Useful links

- [Figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=66%3A881&t=eKMC6v66Cr7VpnHi-0)

ashayp22:
## Description

- Add a back button to `/form/location` on the panel details page
- Hide hamburger on mobile

Fixes #60


components/BackButton.tsx
```diff
@@ -0,0 +1,19 @@
+import { FC } from 'react';
+import { LeftChevron } from './icons/sidebar_icons';
+import { useRouter } from 'next/router';
+
+const BackButton: FC = () => {
+  const router = useRouter();
+
+  return (
+    <div
```
AndrewLester: Make this a button so it's accessible (tabable, screen readers, etc.)
Also replace closing tag
```suggestion
    <button
```

components/Layout.tsx
```diff
@@ -1,15 +1,12 @@
...
-import { useUser } from '@auth0/nextjs-auth0';
 
 const Layout: FC<PropsWithChildren> = ({ children }) => {
-  const { user } = useUser();
-
   return (
     <>
       <NavBar />
-      {user && <Sidebar />}
+      <Sidebar />
```
AndrewLester, line 9: I think this is removing the conditional rendering you added to this file earlier, maybe revoke this.

pages/form/details.tsx
```diff
@@ -37,7 +38,8 @@ const Form = () => {
 
   return (
     <form onSubmit={onSubmit}>
-      <h1 className="font-bold text-4xl mt-8 dark:text-ocf-gray mb-5">
+      <BackButton />
```
AndrewLester, line 41: Move this outside the form so it doesn't automatically submit the form. It's also unrelated semantically so should be separated. I know this might ruin some margins, but then perhaps the margins should be outside the form too.

If you must keep this in the form give the button element `type="button"`

AndrewLester (changes requested):
A few changes, but looks great overall. Also I think the build should be fixed after reverting the changes to `Layout.tsx` (or merge main).

components/Layout.tsx
ashayp22: I removed it because I realized it was unnecessary.

components/icons/sidebar_icons.tsx
```diff
@@ -84,3 +84,20 @@ export const ExitIcon: React.FC = ({}) => (
...
+  <svg
+    width="10"
+    height="28"
+    viewBox="0 0 10 16"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      fill-rule="evenodd"
+      clip-rule="evenodd"
```
AndrewLester: My bad because I didn't catch this on first review. React wants these to be camel case.
```suggestion
      fillRule="evenodd"
      clipRule="evenodd"
```

AndrewLester (approved):
Approving with one thing I missed. Can be suggestion-committed.

PR #86, opened 2023-03-07
Fix preview deployments + build

AndrewLester:
`inverters` will be undefined if there's no logged in user on this page, and they can actually see this page. This won't happen in prod but _will_ happen in preview since we disable auth0.

### DIFF
--- a/pages/account.tsx
+++ b/pages/account.tsx
@@ -44,7 +44,7 @@ const Account: NextPage<Props> = ({ inverters }) => {
           Linking status success: {query.linkSuccess}
         </p>
       )}
-      {inverters?.length ?? 0 > 0 ? (
+      {inverters?.length ? (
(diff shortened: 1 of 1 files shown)


pages/account.tsx
```diff
@@ -44,7 +44,7 @@ const Account: NextPage<Props> = ({ inverters }) => {
           Linking status success: {query.linkSuccess}
         </p>
       )}
-      {inverters.length > 0 ? (
+      {inverters?.length ?? 0 > 0 ? (
```
ericcccsliu: maybe we ditch `??` and just do `inverters?.length` as the logical expression for the ternary here? having `0 > 0` is a bit weird to me. 

```suggestion
      {inverters?.length ? (
```

ericcccsliu (approved):
Looks good, tiny comment

PR #90, opened 2023-03-09
Change Input mode on location details page

code-arman:
## Status:
:rocket: Ready

## Description
Adds pattern and input mode props to input elements so that the numeric keyboard shows up on mobile. This makes it way easier to enter numbers on the location details page


AndrewLester (approved):
Thanks

PR #93, opened 2023-03-09
Fix remaining preview issues

AndrewLester:
## Status:

:rocket: Ready

## Description

Allows Auth0 to work on preview deployments (only deploy preview URLs with commit hash in them).

Preview is not currently showing the hamburger button because a check for a user to exist hides it, when there is no user in preview.


ericcccsliu (approved):
looks good, hopefully this works

PR #98, opened 2023-03-10
Create Sunny Timeframe Module

### LINKED ISSUE #88: Sunny timeframe module
## Description
create a module that tells our users how long it'll be sunny for! 🌤️ 

Screen Shot 2023-03-07 at 12 33 03 AM

this should work off of the same threshold as our `ThresholdGraph` component. 

## Todos

- [ ] write helper function to determine how long we are above the threshold for in our forecast 
- [ ] display that data in module 
- [x] style according to figma 

## Useful links

- [figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=188%3A1116&t=egs2SijR0fLl8xgd-1)

## Time estimate
tbd!

ashayp22:
## Description

Create a module that tells our users how long it'll be sunny for! 

Fixes #88


components/dashboard_modules/SunnyTimeframe.tsx
```diff
@@ -0,0 +1,44 @@
...
+  graphThreshold,
+  Value,
+} from 'lib/utils';
+
+const SunnyTimeframe: FC = () => {
+  const { data } = useFutureGraphData();
+  const nextThreshold = getNextThresholdIndex(
+    data?.forecast_values,
+    graphThreshold
+  );
```
AndrewLester: ~This computation isn't trivial, so while I wouldn't normally recommend it, maybe wrap this in a `useMemo`.~

components/dashboard_modules/SunnyTimeframe.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  if (nextThreshold) {
```
AndrewLester: Instead of putting this all inside the `if`, maybe do an `if (!nextThreshold)` with an early `return null`.

lib/utils.ts
```diff
@@ -133,19 +133,108 @@ export const getArrayMaxOrMinAfterIndex = (
...
+
+    const value =
+      forecast_values[currentIndex].expected_generation_kw >= threshold
+        ? Value.Max
+        : Value.Min;
+
+    while (currentIndex < forecast_values.length) {
+      if (
+        operator *
+          (forecast_values[currentIndex].expected_generation_kw - threshold) >
```
AndrewLester: Move this computation to above the `if` and call it `thresholdDifference` or something?
```suggestion
        thresholdDifference >
```

lib/utils.ts
```diff
@@ -133,19 +133,108 @@ export const getArrayMaxOrMinAfterIndex = (
...
+ * the sunny threshold
+ * @param forecast_values expected generated forecast values (kilowatts) at specific times
+ * @param threshold sunny threshold in kilowatts
+ * @returns Object containing hour difference between the next date and
+ * if this date is below (Value.Max) or above (Value.Min) the threshold
+ */
+export const getNextThresholdIndex = (
+  forecast_values: ForecastDataPoint[] | undefined,
+  threshold: number
+): MinMaxInterface | null => {
```
AndrewLester: I'm not sure I love using `MinMaxInterface`'s value (`Value`) to represent "above" or "below" the threshold. Maybe just return an object with the type `{ aboveThreshold: boolean, value: number }`

lib/utils.ts
```diff
@@ -133,19 +133,108 @@ export const getArrayMaxOrMinAfterIndex = (
...
+        ? Value.Max
+        : Value.Min;
+
+    while (currentIndex < forecast_values.length) {
+      if (
+        operator *
+          (forecast_values[currentIndex].expected_generation_kw - threshold) >
+        0
+      ) {
+        console.log(currentIndex);
```
AndrewLester:
```suggestion

```

lib/utils.ts
```diff
@@ -133,19 +133,108 @@ export const getArrayMaxOrMinAfterIndex = (
...
+/**
+ * Determines the hour difference between the current time and the next time we are above or below
+ * the sunny threshold
+ * @param forecast_values expected generated forecast values (kilowatts) at specific times
+ * @param threshold sunny threshold in kilowatts
+ * @returns Object containing hour difference between the next date and
+ * if this date is below (Value.Max) or above (Value.Min) the threshold
+ */
+export const getNextThresholdIndex = (
+  forecast_values: ForecastDataPoint[] | undefined,
```
AndrewLester: Require this parameter be defined and remove the check within the function. This might be me, but I prefer doing checks for existence outside core, computational functions that deal with transforming data.
```suggestion
  forecast_values: ForecastDataPoint[],
```

lib/utils.ts
```diff
@@ -133,19 +133,108 @@ export const getArrayMaxOrMinAfterIndex = (
...
+
+    const operator =
+      forecast_values[currentIndex].expected_generation_kw >= threshold
+        ? -1
+        : 1;
+
+    const value =
+      forecast_values[currentIndex].expected_generation_kw >= threshold
+        ? Value.Max
+        : Value.Min;
```
AndrewLester: If switching from `MinMaxInterface`, this'll be more like:
```suggestion
    const aboveThreshold = forecast_values[currentIndex].expected_generation_kw >= threshold;
```

components/dashboard_modules/SunnyTimeframe.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    return (
+      <div className="flex-1 my-2 p-4 text-center bg-ocf-gray-1000 rounded-2xl">
+        <div className="mb-2 text-xs text-ocf-gray font-semibold">
+          {sunnyText}
+        </div>
+        <div className="mb-1 text-2xl text-ocf-yellow font-semibold">
+          {number} {timeUnit}
+        </div>
+      </div>
```
AndrewLester: Can you use a [`<NumberDisplay`](https://github.com/openclimatefix/pv-sites-mobile/blob/main/components/dashboard_modules/NumberDisplay.tsx) for this?

AndrewLester (changes requested):
Time unit / display logic is very clean, and thanks for moving `getCurrentTimeForecastIndex`.

A few changes within the computation and some code reuse things. Also maybe we don't need `useMemo` actually, this component isn't really going to rerender often.

components/dashboard_modules/SunnyTimeframe.tsx
AndrewLester: This component won't rerender much I don't think, so hold off on this for now.

data/pv-forecast.json
```diff
@@ -1,259 +1,259 @@
 {
   "forecast_uuid": "95833950-de0f-4447-8c73-ffa5138b225d",
   "site_uuid": "b97f68cd-50e0-49bb-a850-108d4a9f7b7e",
-  "forecast_creation_datetime": "2023-04-16T01:26:20.557211+00:00",
+  "forecast_creation_datetime": "2024-04-16T01:26:20.557211+00:00",
```
AndrewLester, line 4: Is there any reason for bumping these years up?

AndrewLester (approved):
Great work

ericcccsliu (approved):

data/pv-forecast.json
ashayp22: So the Sunny Timeframe module displays a "Sunny in..." message.

AndrewLester: Gotcha

PR #100, opened 2023-03-11
Fix logout redirect

AndrewLester:
This fixes the logout page on preview and production deployments.


ericcccsliu (approved):
Great, hope this works!

PR #101, opened 2023-03-11
Create custom time hook

### LINKED ISSUE #97: Extract time-related logic into custom hook
## Description
Currently, we're synchronizing our threshold graph as well as some of our other dashboard modules with the current time. In the threshold graph, we have the following code: 

https://github.com/openclimatefix/pv-sites-mobile/blob/8fa566ef109b31145439fda3ac634f49ef7ce0c9/components/graphs/ThresholdGraph.tsx#L31-L40

We'd like to extract this into a custom hook, perhaps called `useTime` or something similar. Additionally, we'd like to refactor all other instances of utilizing the current time to use our new `useTime` hook. 

Furthermore, in the future, some of our modules will benefit from knowing whether it is day or night, in order to be able to display the most relevant information at all times. To this end, we should be able to use the [SunCalc](https://www.npmjs.com/package/suncalc) library, in conjunction with our users' latitude and longitude, to get times for sunrise and sunset (and dusk, dawn, etc...). We'd like to use the SunCalc library to add a `isNighttime` state to our custom hook, so we may inform modules when it is nighttime and have them update accordingly. 

## Todos

- [x] refactor interval logic into a custom hook. **Consider:** Should the custom hook do formatting internally, or should that be left up to _consumers_ of the hook?
- [x] refactor files relying on current time to use the current time returned by our custom hook 
- [x] use the SunCalc library to determine whether it is day or night 
- [x] add a state to our time hook to allow components to find out whether it is day or night 

## Useful links

- [intro to react custom hooks](https://beta.reactjs.org/learn/reusing-logic-with-custom-hooks)
- [suncalc](https://www.npmjs.com/package/suncalc)

## Time estimate
tbd!

rohanvan123:
## Description
For this issue, I was assigned to abstract the logic that was used in the threshhold graph for calculating the current time into its own hook, which is called `useTime()`. This hook returns the currentTime state variable to access the current user's time based on his/her timezone. It also makes use of the SunCalc library, which it utilized to determine whether or not it is daylight or nighttime based on comparing the current time to sunrise and sunset times. The hook also returns a state variable isDaytime (the sunrise sunset times are based in london) to access this information elsewhere. 

Fixes #97 

## Todos
- Code formatting
- Ensure that hook works as intended for the SunCalc Library

### DIFF
--- a/components/graphs/ThresholdGraph.tsx
+++ b/components/graphs/ThresholdGraph.tsx
@@ -41,7 +41,7 @@ const ThresholdGraph = () => {
   //   return () => clearInterval(intervalId);
   // });
 
-  const currentTime = useTime();
+  const [currentTime, isDaytime] = useTime();
 
   /**
    * Renders a text label for the threshold
--- a/components/graphs/ThresholdGraph.tsx
+++ b/components/graphs/ThresholdGraph.tsx
@@ -25,19 +25,23 @@ import {
 } from 'lib/utils';
 
 import { useFutureGraphData } from 'lib/hooks';
+import useTime from '~/lib/hooks/useTime';
 
 const ThresholdGraph = () => {
   const { data, isLoading } = useFutureGraphData();
-  const [currentTime, setCurrentTime] = useState(formatter.format(Date.now()));
+  // const [currentTime, setCurrentTime] = useState(formatter.format(Date.now()));
 
-  useEffect(() => {
-    const intervalId = setInterval(() => {
-      setCurrentTime(formatter.format(Date.now()));
-    }, 1000);
+  // useEffect(() => {
+  //   const intervalId = setInterval(() => {
+  //     setCurrentTime(formatter.format(Date.now()));
+  //     console.log(currentTime);
+  //   }, 1000);
 
-    // clear interval on re-render to avoid memory leaks
-    return () => clearInterval(intervalId);
-  });
+  //   // clear interval on re-render to avoid memory leaks
+  //   return () => clearInterval(intervalId);
+  // });
+
+  const currentTime = useTime();
 
   /**
    * Renders a text label for the threshold
--- a/components/graphs/ThresholdGraph.tsx
+++ b/components/graphs/ThresholdGraph.tsx
@@ -29,7 +29,7 @@ import useTime from '~/lib/hooks/useTime';
 
 const ThresholdGraph = () => {
   const { data, isLoading } = useFutureGraphData();
-  const { currentTimeNoFilter, isDaytime } = useTime();
+  const { currentTimeNoFilter, isDaytime } = useTime(51.5, -0.1);
 
   /**
    * Renders a text label for the threshold
--- a/components/graphs/ThresholdGraph.tsx
+++ b/components/graphs/ThresholdGraph.tsx
@@ -29,7 +29,7 @@ import useTime from '~/lib/hooks/useTime';
 
 const ThresholdGraph = () => {
   const { data, isLoading } = useFutureGraphData();
-  const { currentTimeNoFilter, isDaytime } = useTime(51.5, -0.1);
+  const { currentTime, isDaytime } = useTime(51.5, -0.1);
 
   /**
    * Renders a text label for the threshold
@@ -210,7 +210,7 @@ const ThresholdGraph = () => {
         suppressHydrationWarning
         className="text-white text-base font-semibold"
       >
-        {formatter.format(currentTimeNoFilter)}
+        {formatter.format(currentTime)}
       </p>
     );
   };
--- a/components/graphs/ThresholdGraph.tsx
+++ b/components/graphs/ThresholdGraph.tsx
@@ -29,19 +29,7 @@ import useTime from '~/lib/hooks/useTime';
 
 const ThresholdGraph = () => {
   const { data, isLoading } = useFutureGraphData();
-  // const [currentTime, setCurrentTime] = useState(formatter.format(Date.now()));
-
-  // useEffect(() => {
-  //   const intervalId = setInterval(() => {
-  //     setCurrentTime(formatter.format(Date.now()));
-  //     console.log(currentTime);
-  //   }, 1000);
-
-  //   // clear interval on re-render to avoid memory leaks
-  //   return () => clearInterval(intervalId);
-  // });
-
-  const [currentTime, isDaytime] = useTime();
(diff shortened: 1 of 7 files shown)


lib/hooks/useTime.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useState } from 'react';
+import { formatter } from '../utils';
+
+const useTime = () => {
+  const SunCalc = require('suncalc');
```
AndrewLester: Try moving this line to the top of the file and change the `require` syntax into standard `import` syntax.

lib/hooks/useTime.ts
```diff
@@ -0,0 +1,35 @@
...
+  const [currentTime, setCurrentTime] = useState(formatter.format(Date.now()));
+  const [currentTimeNoFilter, setCurrentTimeNoFilter] = useState(Date.now());
```
AndrewLester: Perhaps this `useTime` hook should deal solely with "no filter" time, or unformatted time. This way, it can be most easily used by the largest possible group of consumers.

lib/hooks/useTime.ts
```diff
@@ -0,0 +1,35 @@
...
+  const [isDaytime, setIsDaytime] = useState(false);
```
AndrewLester: Since this stats variable is set upon every update to `currentTimeNoFilter` and is strictly dependent on that value, and that value itself will update every render due to the useEffect, perhaps you don't need this state variable.

Instead, you can opt to compute `isDaytime` right here in the body of the hook, and simply return it in the list at the bottom.

lib/hooks/useTime.ts
```diff
@@ -0,0 +1,35 @@
...
+      } else {
+        setIsDaytime(true);
+      }
+    }, 1000);
+
+    // clear interval on re-render to avoid memory leaks
+    return () => clearInterval(intervalId);
+  });
+
+  return [currentTime, isDaytime];
```
AndrewLester: While the `useState` hook does return an array of stuff like this, it's customary to return an object with properties and values in a custom hook. This is just generally more readable at the consumer side.


```suggestion
  return { currentTime, isDaytime} ;
```

components/graphs/ThresholdGraph.tsx
```diff
@@ -25,19 +25,23 @@ import {
...
+  //   }, 1000);
 
-    // clear interval on re-render to avoid memory leaks
-    return () => clearInterval(intervalId);
-  });
+  //   // clear interval on re-render to avoid memory leaks
+  //   return () => clearInterval(intervalId);
+  // });
+
+  const [currentTime, isDaytime] = useTime();
```
AndrewLester: After you make the swap from list &rarr; object I mention above:
```suggestion
  const { currentTime, isDaytime } = useTime();
```

components/graphs/ThresholdGraph.tsx
```diff
@@ -25,19 +25,23 @@ import {
 } from 'lib/utils';
 
 import { useFutureGraphData } from 'lib/hooks';
+import useTime from '~/lib/hooks/useTime';
 
 const ThresholdGraph = () => {
   const { data, isLoading } = useFutureGraphData();
-  const [currentTime, setCurrentTime] = useState(formatter.format(Date.now()));
+  // const [currentTime, setCurrentTime] = useState(formatter.format(Date.now()));
```
AndrewLester: All these commented lines can be removed before merge (this component becomes significantly simpler)

AndrewLester (changes requested):
Great work, this should make the components we write that need to check for daytime way simpler. 

Only a few small changes.

lib/hooks/useTime.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useState } from 'react';
+import { formatter } from '../utils';
+
+const useTime = () => {
+  const SunCalc = require('suncalc');
+  const [currentTime, setCurrentTime] = useState(formatter.format(Date.now()));
+  const [currentTimeNoFilter, setCurrentTimeNoFilter] = useState(Date.now());
+  const [isDaytime, setIsDaytime] = useState(false);
+  // get sunrise/sunset info for london
+  const times = SunCalc.getTimes(Date.now(), 51.5, -0.1);
```
ericcccsliu: can we pass in lat, long to the useTime hook? then, wherever we use the hook, we can pass in the lat, long of the relevant site.

components/graphs/ThresholdGraph.tsx
ericcccsliu: make sure to remove comments before we can merge!

ericcccsliu (changes requested):
great work getting this PR out so quickly!

lib/hooks/useTime.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Creates a hook more accessesing specific values like the current time, whether it is daytime
+ * @param latitude, the latitude float value that is passed in
+ * @param longitude, the longitude number value that is passed in
+ * @returns currentTimeNoFilter, unfiltered version of the current time (right now) that a user can format
+ * @returns isDaytime, boolean value indicating whether it is daytime or not based on current time zone.
+ */
+const useTime = (latitude: number, longitude: number) => {
+  const [currentTimeNoFilter, setCurrentTimeNoFilter] = useState(Date.now());
```
ericcccsliu: don't think we need to specify `NoFilter` here—seems like we can just call these`[currentTime, setCurrentTime]`

lib/hooks/useTime.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  // get sunrise/sunset time fors passed in location
```
ericcccsliu:
```suggestion
  // get sunrise/sunset time for passed in location
```

ericcccsliu (approved):
great work!!!

lib/utils.ts
```diff
@@ -226,3 +227,11 @@ export const getNextThresholdIndex = (
...
 /* Represents the threshold for the graph */
 export const graphThreshold = 0.7;
+
+/**
+ * Custom hook for time
+ * @param array Array to search for a local minimum or maximum value
+ * @param key The key that represents the values being compared in array
+ * @param startIndex The index to start the search at
+ * @returns The index of the next minimum or maximum value
+ */
```
AndrewLester: Can these comments at the bottom here be removed?

```suggestion

```

AndrewLester (approved):
One last small thing, really nice work.

PR #104, opened 2023-03-12
Add Reverse Geocoding

ashayp22:
## Description

Adds reverse geocoding to the Site Location map. Whenever a user moves the map, the latitude/longitude of the center of the map is converted into an address, which is displayed in the MapBox Geocoder (search bar).

## Todos

- [x] The map zooms into a location whenever the user moves the marker, which makes it harder to zoom out. Need to only apply reverse geocoding when fully zoomed in.


components/LocationInput.tsx
```diff
@@ -14,21 +14,26 @@ interface LocationInputProps {
   setLngExternal: (lng: number) => void;
   setLatExternal: (lat: number) => void;
   zoomLevelThreshold: number;
+  reverseGeocodingZoomThreshold: number;
 }
 
+const originalLng = -2.3175601;
+const originalLat = 54.70534432;
```
AndrewLester, line 20: Thanks for moving these

components/LocationInput.tsx
```diff
@@ -80,18 +89,30 @@ const LocationInput: FC<PropsWithChildren<LocationInputProps>> = ({
...
+          geocoder.query(`${savedLat}, ${savedLng}`);
+        }
+
+        return setIsSubmissionEnabled(currentZoom > zoomLevelThreshold);
+      });
     }
   });
 
+  useEffect(() => {});
+
```
AndrewLester:
```suggestion

```

components/LocationInput.tsx
```diff
@@ -80,18 +89,30 @@ const LocationInput: FC<PropsWithChildren<LocationInputProps>> = ({
...
+        const currentZoom = map.current!.getZoom();
+
+        if (currentZoom >= reverseGeocodingZoomThreshold) {
+          geocoder.query(`${savedLat}, ${savedLng}`);
+        }
+
+        return setIsSubmissionEnabled(currentZoom > zoomLevelThreshold);
+      });
     }
   });
```
AndrewLester: I'm now realizing this `useEffect` doesn't declare a dependency array of `[]`... this is probably not what we want. I see this was here before, and everything does work. Maybe in the future we fix this. I could make an issue or you could try with it.

AndrewLester (changes requested):
Looks great, will leave more comments tomorrow. Two things:
- Maybe to improve the dragging UX you could use some setting of [setFlyTo](https://github.com/mapbox/mapbox-gl-geocoder/blob/main/API.md#setflyto) so that the drag doesn't lock into a specific point on the map (since we look for specific lat/lng locking to points of interest isn't important). We DO want `flyTo` enabled on search though.
- What's up with the yarn.lock change? Seems like it removed quotes? Is this a new yarn update?

ashayp22:
@AndrewLester I resolved your code review comments. I added [setFlyTo](https://github.com/mapbox/mapbox-gl-geocoder/blob/main/API.md#setflyto) functionality so that it is enabled on search and disabled when performing the reverse geocoding.

components/LocationInput.tsx
```diff
@@ -70,7 +74,17 @@ const LocationInput: FC<PropsWithChildren<LocationInputProps>> = ({
 
       map.current.on('load', () => setIsMapReady(true));
 
-      map.current.on('movestart', () => setIsSubmissionEnabled(false));
+      map.current.on('idle', () => {
+        // Enables fly to animation on search
+        geocoder.setFlyTo(true);
+      });
```
AndrewLester, line 80: Great idea putting this in 'idle'.

AndrewLester (approved):
Nice work setting this up. I'm getting the yarn.lock change locally too now, and after a google it seems like it was related to the package-lock.json being in here... this is a fix it seems.

PR #106, opened 2023-03-13
Implement power requirement dashboard module

### LINKED ISSUE #87: Power requirement dashboard module
## Description
Create a module on our dashboard that displays common appliances that a user may utilize solely using the current forecasted output of their solar panels. 

Screen Shot 2023-03-07 at 12 21 28 AM

Consider storing the content to be displayed in a JSON file, similar to issue #66. Consult @anyaparekh on how this should be done. Ideally, the same JSON file would be used to store content for both this issue and #66. 

## Todos
- [x] create module
- [x] add relevant content to JSON file 
- [x] display relevant content based on current predicted output
- [x] style according to figma 
- [x] place onto dashboard page 

## Useful links

- [figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=188%3A1116&t=egs2SijR0fLl8xgd-1)

## Time estimate
tbd!

rohanvan123:
## Description
For this issue I abstracted out some of Anya's code for building the carousel component into a dashboard module to display the best appliance recommendation whose energy requirement is less than or equal to the current output. Essentially, this component first retrieved the currentOutput. If this value is undefined, it returns a `<NumberDisplay>` component with value set to "N/A". Otherwise, it finds the index of the appliances list, retrieved from the json file, that has the greatest kW requirement less than or equal to the currentOutput. I created a helper method `getBestRecommendationIndex(currentOutput)` to find the best index accordingly. 

Fixes #87 

## Todos
-Code Quality check
-Ensure that designs are applicable for edge cases

## Screenshots

Screenshot 2023-03-13 at 5 32 53 PM
Screenshot 2023-03-13 at 5 33 20 PM
Screenshot 2023-03-13 at 5 33 48 PM


components/dashboard-modules/EnergyReccomendation.tsx
```diff
@@ -0,0 +1,71 @@
...
+          text-center
+          bg-ocf-gray-1000
+          rounded-2xl"
+      >
+        <Image
+          src={appliance.icon}
+          alt={appliance.name}
+          width={34.5}
+          height={28.75}
+          className=""
```
ericcccsliu: remove this line

components/dashboard-modules/EnergyReccomendation.tsx
```diff
@@ -0,0 +1,71 @@
...
+    for (let i = 0; i < content.appliances.length; i++) {
+      const currAppliance = content.appliances[i];
+      if (
+        Number(currAppliance.kW) > maxKW &&
+        Number(currAppliance.kW) <= currentOutput
+      ) {
+        maxIndex = i;
+        maxKW = Number(currAppliance.kW);
+      }
+    }
```
ericcccsliu: this is great, and should work well, but we can also replace this with a `forEach` function! 

```suggestion
    content.appliances.forEach((appliance, i) => {
        if (
          Number(appliance.kW) > maxKW &&
          Number(appliance.kW) <= currentOutput
        ) {
          maxIndex = i;
          maxKW = Number(appliance.kW);
        }
      }
    )
```
something like this should work (the suggestion I gave might be a bit buggy), and getting into the habit of using the `forEach()` function (and other array prototype functions when applicable) will allow our code to be generally a bit less verbose.

components/dashboard-modules/EnergyReccomendation.tsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+  return -1;
+};
+
+const EnergyRecommendation: FC<{ siteUUID: string }> = ({ siteUUID }) => {
+  const { forecastData, isLoading } = useSiteData(siteUUID);
+  const currentOutput = forecastData
+    ? getCurrentTimeForecast(forecastData.forecast_values)
+    : undefined;
+  const reccomendationIdx = getBestRecomendationIndex(currentOutput);
```
ericcccsliu: a few typos here
```suggestion
  const recommendationIdx = getBestRecommendationIndex(currentOutput);
```
also, would probably be good to just have `recommendationIndex` defined as a ternary 
```suggestion
  const recommendationIdx = currentOutput ? getBestRecommendationIndex(currentOutput) : undefined
```
this will allow us to avoid having to check for undefinedness in getBestRecommendationIndex, because I feel like we shouldn't have to worry about checking for undefinedness in that function.

components/dashboard-modules/EnergyReccomendation.tsx
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Determines the appliance with the greatest energy required that is less than or equal to the current output
+ * @param currentOutput the current output in kW based on the getCurrentTimeForecast
+ * @returns the index of the appliance with the greatest enrgy requiement less than or equal to the current ouput
+ * or -1 if an index could not be found
+ */
+
+const getBestRecomendationIndex = (currentOutput: number | undefined) => {
+  if (currentOutput != undefined) {
+    let maxIndex = -1;
```
ericcccsliu: let's have this be initialized to `undefined`, and have the function return `undefined` when there doesn't exist a suitable index

ericcccsliu: I just don't really like having sentinel numbers here like `-1`; if we don't have a match, returning `undefined` (or even `null`, `null` might be better) seems more honest

ericcccsliu (changes requested):
Absolutely amazing work. Just some code quality improvements (we've got time, anyways) and we'll be good to merge this in.

components/dashboard-modules/EnergyRecommendation.tsx
```diff
@@ -0,0 +1,74 @@
...
+import NumberDisplay from './NumberDisplay';
+
+/**
+ * Determines the appliance with the greatest energy required that is less than or equal to the current output
+ * @param currentOutput the current output in kW based on the getCurrentTimeForecast
+ * @returns the index of the appliance with the greatest enrgy requiement less than or equal to the current ouput
+ * or -1 if an index could not be found
+ */
+
+const getBestRecommendationIndex = (currentOutput: number | undefined) => {
```
ericcccsliu: we should be able to remove `undefined` and the `if (currentOutput != undefined)` check in this function because it should never be used when `currentOutput == undefined`

components/dashboard-modules/EnergyRecommendation.tsx
```diff
@@ -0,0 +1,74 @@
...
+    return (
+      <div
+        className="
+          flex-1
+          flex
+          my-2
+          p-4
+          text-center
+          bg-ocf-gray-1000
+          rounded-2xl"
```
ericcccsliu: could these styles be changed a bit to keep the content within the module centered as screen size increases? 

Screen Shot 2023-03-13 at 9 09 00 PM

ericcccsliu (approved):
Very small changes needed, feel free to merge once you think you're done.

components/dashboard-modules/EnergyRecommendation.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Determines the appliance with the greatest energy required that is less than or equal to the current output
+ * @param currentOutput the current output in kW based on the getCurrentTimeForecast
+ * @returns the index of the appliance with the greatest enrgy requiement less than or equal to the current ouput
+ * or -1 if an index could not be found
+ */
+
+const getBestRecommendationIndex = (currentOutput: number) => {
+  if (currentOutput != undefined) {
```
AndrewLester: No need to check for undefined here since `currentOutput` is of type `number` (which excludes `undefined` unless included explicitly). That's the power of TypeScript™.

components/dashboard-modules/EnergyRecommendation.tsx
```diff
@@ -0,0 +1,75 @@
...
+const EnergyRecommendation: FC<{ siteUUID: string }> = ({ siteUUID }) => {
+  const { forecastData, isLoading } = useSiteData(siteUUID);
+  const currentOutput = forecastData
+    ? getCurrentTimeForecast(forecastData.forecast_values)
+    : undefined;
+
+  // const recommendationIdx = currentOutput
+  //   ? getBestRecommendationIndex(currentOutput)
+  //   : null;
+
```
AndrewLester: Can remove commented out code.
```suggestion
```

components/dashboard-modules/EnergyRecommendation.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const recommendationIdx = getBestRecommendationIndex(5);
```
AndrewLester: Is this supposed to be changed to this?
```suggestion
  const recommendationIdx = getBestRecommendationIndex(currentOutput);
```

AndrewLester (approved):
Very nice work, the dashboard is looking clean

PR #111, opened 2023-03-14
Switch to new api urls

AndrewLester:
## Status:

:rocket: Ready

## Description

Switch to new API urls and add aggregate URLs to mock routes.


ericcccsliu (approved):
LGTM

PR #114, opened 2023-03-14
Aggregate predictions over multiple sites

### LINKED ISSUE #89: Aggregate predictions over multiple sites
## Description

For our aggregate dashboard over multiple sites, we need aggregate metrics of everything we currently want to display. To this end, we'll need to write functions to aggregate the data returned by the API. Specifically, we'll want to create a new forecast, made up of all the individual site forecasts aggregated together. We'll also want to aggregate together capacity from all the sites, so that we can return the percentage capacity outputted at any moment over all solar sites. 

With this aggregated forecast and aggregated capacity, we should have all the data we need to pass into our modules and have them work as expected. 

~Should not be addressed until [this pv-site-api PR is merged in](https://github.com/openclimatefix/pv-site-api/pull/50)~, now merged

You can use their new endpoint (which we have mock endpoints for) `/sites/pv_forecast?site_uuids=uuid_1,uuid_2` to get several site's forecasts at once.

## Todos 

- [x] aggregate pv forecasts over all sites 
- [x] aggregate capacities over all sites 

## Time estimate
tbd

ashayp22:
## Description

Create the `useSiteAggregation()` hook to aggregate predictions over multiple sites, including total solar energy generated at specific date-times and the sum of solar panel capacity across all sites.

Fixes #89 

## Todos

- [x] Create hook that aggregates PV forecasts and capacities over all sites
- [x] Test the hook
- [x] Add any additional necessary aggregations to the hook

### DIFF
--- a/lib/hooks/useSiteAggregation.ts
+++ b/lib/hooks/useSiteAggregation.ts
@@ -18,16 +18,14 @@ const useSiteAggregation = (allSiteUUID: string[]) => {
     siteListFetcher
   );
 
-  const parsedAllSiteUUID = allSiteUUID
-    .reduce((prev, curr) => `${prev},${curr}`)
-    .substring(1);
-
   const {
     data: manyForecastData,
     error: manyForecastError,
     isLoading: isManyForecastLoading,
   } = useSWR(
-    `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/pv_forecast?site_uuids=${parsedAllSiteUUID}`,
+    `${
+      process.env.NEXT_PUBLIC_API_BASE_URL
+    }/api/pv_forecast?site_uuids=${allSiteUUID.join(',')}`,
     manyForecastDataFetcher
   );
 
@@ -41,8 +39,9 @@ const useSiteAggregation = (allSiteUUID: string[]) => {
           prevTotalSiteCapacity + newSite.installed_capacity_kw,
         0
       )
-    : 0;
-  let totalExpectedGeneration: ForecastDataPoint[] = [];
+    : undefined;
+
+  let totalExpectedGeneration: ForecastDataPoint[] | undefined = undefined;
 
   // Sum the expected generation for each date returned in all of the site forecasts
   if (manyForecastData) {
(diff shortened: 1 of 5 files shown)


ashayp22:
I added a sum aggregation to the site capacity and sum aggregation based on date-time to the solar forecasts. Please let me know if this is sufficient for the dashboard, and if not, what I should change.

lib/hooks/useSiteAggregation.ts
```diff
@@ -0,0 +1,93 @@
...
+  const isLoading = isSiteListLoading || isManyForecastLoading;
+
+  // Sum the installed capacity at all of the sites
+  const totalInstalledCapacityKw = siteListData
+    ? siteListData.site_list.reduce(
+        (prevTotalSiteCapacity, newSite) =>
+          prevTotalSiteCapacity + newSite.installed_capacity_kw,
+        0
+      )
+    : 0;
```
ericcccsliu: maybe we should be treating this as `undefined` rather than 0 when siteListData is falsy?

lib/hooks/useSiteAggregation.ts
```diff
@@ -0,0 +1,93 @@
...
+    error: siteListError,
+    isLoading: isSiteListLoading,
+  } = useSWR(
+    `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/sites`,
+    siteListFetcher
+  );
+
+  const parsedAllSiteUUID = allSiteUUID
+    .reduce((prev, curr) => `${prev},${curr}`)
+    .substring(1);
```
AndrewLester: .join(','); will also work here.

ericcccsliu (approved):

AndrewLester (approved):
Great work, just one small thing.

ericcccsliu (commented):
Since the total capacity and the forecast data are there, this should be good for our dashboard for now. I think we should be building out a slightly different dashboard page for aggregate data, so we don't really need to return _all_ of the info that useSiteData returns, and this is all we need for graphs, current output, and current capacity modules to work.

PR #119, opened 2023-03-15
Add automatic revalidation interval

ericcccsliu:
## Status:
:rocket: Ready

## Description
Just one change 
```TypeScript
<SWRConfig value={{ fetcher: fetcher, refreshInterval: 10000 }}>
```
adding a refreshInterval parameter to our SWRConfig, so by default, all our useSWR calls will revalidate after 10 seconds. We can also easily increase this value to have less unnecessary calls to our API.


AndrewLester (approved):
This is super cool, I think we just wanna be careful if using their actual api to avoid hitting it too much.

PR #122, opened 2023-03-16
Fix site page keyboard navigation

### LINKED ISSUE #121: Site cards not keyboard focusable in site page
**Describe the bug**
In the site page, the cards (which are `Link` components) are not reachable by keyboard navigation 

**To Reproduce**

1. go to site page
2. try to keyboard focus on cards 
3. you can't :( 

**Expected behavior**
cards should be keyboard-focusable

ericcccsliu:
## Status:
:rocket: Ready

## Description

- Issue with the keyboard nav was that the child of the `Link` in the site card was not an `<a>` tag but rather a `<span>`. Element only registers as a hyperlink if it is an `<a>` tag. 
  - to fix this, I simply wrapped the `Link` around `<a>` and use the `passHref` prop to force the `Link` to pass its href to its child 
- We also needed to disable the link functionality of the card component in edit mode. Because the href property of a `Link` cannot be omitted, I refactored everything into the card into a new component, and wrapped that component into a `Link` conditionally
- Also wrapped the icons for edit and delete in `<button>` attributes, allowing them to be tabbable 

Fixes #121 

## Screenshots
Screen Shot 2023-03-16 at 2 21 38 PM


pages/sites.tsx
```diff
@@ -18,10 +18,10 @@ const Sites = () => {
...
         </button>
       </div>
-      <SiteCard isEditMode={editMode} />
-      <SiteCard isEditMode={editMode} />
-      <SiteCard isEditMode={editMode} />
-      <SiteCard isEditMode={editMode} />
+      <SiteCardLink isEditMode={editMode} />
+      <SiteCardLink isEditMode={editMode} />
+      <SiteCardLink isEditMode={editMode} />
+      <SiteCardLink isEditMode={editMode} />
```
ericcccsliu, line 24: passing `editMode` into all of these is a bit not fun—thoughts? hook this?

AndrewLester (approved):
I like the separation here between link and the other component. For now this is fine, but in the future I think we'll ideally want to move the link deeper into the component and only wrap the parts of it that aren't buttons.

Two reasons for this:
- don't have to remove the link in edit mode
- children of an <a aren't supposed to be interactive by HTML standards (so no <button as children for example)

PR #125, opened 2023-03-17
Fix flashing navbar and hamburger button

### LINKED ISSUE #118: Bottom navbar flashes on reload with empty sitelist
**Describe the bug**
Upon refresh, with no sites, the bottom navbar briefly appears then disappears 

**To Reproduce**

1. Go to the form details page (any other page with bottom nav should display the same) 
2. Ensure there are no sites in the site list in our mock data
3. Refresh
4. See bottom navbar briefly flash in and out of view 

**Expected behavior**
No flashing of the bottom navbar should occur

**Additional context**
This problem was fixed for our menu icon by hiding it by default and fading it in if sites were present in the site list. We can consider fixing this issue with CSS transitions or consider fixing it by different means altogether

AndrewLester:
## Status:

:rocket: Ready

## Description

Uses `getInitialProps` in `_app.tsx` to load user's sites on server, allowing UI to load before first client render.

Fixes #118


pages/_app.tsx
```diff
@@ -1,20 +1,39 @@
...
-const App: FC<AppProps> = ({ Component, pageProps }) => {
+type InitialProps = { siteList?: SiteList };
+
+type SitesAppType = NextComponentType<
+  AppContext,
+  InitialProps,
+  AppProps<{ user: UserProfile }> & InitialProps
+>;
+
+const SitesApp: SitesAppType = ({ Component, pageProps, siteList }) => {
```
ericcccsliu: I'm thinking that `SitesApp` might not be the best name for this—why not just keep it as `App`? It does retrieve sites on server side, but it also still controls everything else.

ericcccsliu (approved):

PR #126, opened 2023-03-17
Add Desktop Support for Site Details Page 

### LINKED ISSUE #99: Location input page desktop design changes
## Description

The desktop designs for the location input page have recently changed, and we need to independently make the form look different on desktop than on mobile.

image

[Figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=0%3A1&t=MB65jb5aQPcvqULY-0)

## Todos

- [ ] Since the desktop and mobile designs look so different, for now render a different component on desktop to avoid messing with the correct mobile HTML/tailwind. Detect mobile/desktop with `sm:block hidden`, for example, which will only display on desktop. Use the Tailwind SM breakpoint (or whatever the breakpoint for bottom nav is) for any media query, though you might want to wrap the current mobile stuff in the opposite (`sm:hidden block`). You may be able to reuse parts of the [`LocationInput`](https://github.com/openclimatefix/pv-sites-mobile/blob/main/components/LocationInput.tsx) mapbox code, so try to extract parts of it into a reusable point.
- [ ] Setup first screen with search bar
- [ ] Setup searchbar using mapbox geocoder. If there's no search, the "suggested location" should say "Current Location" and clicking it should [trigger find my location](https://docs.mapbox.com/mapbox-gl-js/api/markers/#geolocatecontrol#trigger). [Geocoder docs](https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder/) - [more Geocoder docs](https://github.com/mapbox/mapbox-gl-geocoder/blob/main/API.md#mapboxgeocoder).
- [ ] Searching should put marker on map and zoom in.
- [x] Dragging the map moves the marker (or keeps the marker in the center of the screen, if you will). This is current behavior. You'll need to then use reverse geocoding to lookup the lat, lng of the marker's position to change the text in the search, now pinpointed location, bar. See links. Clicking the pinpointed location bar in this stage goes back to the search stage, and focuses the input.
- [ ] Clicking "Confirm", moves to the next page
- [ ] Clicking the usual "find my location" button should also make the marker appear at that location.

## Useful links

- [Designs](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=0%3A1&t=MB65jb5aQPcvqULY-0)
- https://github.com/mapbox/mapbox-gl-geocoder/blob/main/API.md#transformfeaturetogeolocationtext - could be useful
- [reverse geocoding](https://docs.mapbox.com/api/search/geocoding/#reverse-geocoding)

## Time estimate
tbd

ashayp22:
## Description

The desktop designs for the location input page have recently changed, and we need to independently make the form look different on desktop than on mobile. This pull request adds the following desktop changes:

- Refactor Form Context by saving more information about a solar site.
- Add a popup to the map after a user marks a location.
- Implement Location and Form pages for the desktop view.
- Add a site name input field.

Fixes #99 

## Todos

- [x] Allow user input to be saved when navigating the form
- [x] Add desktop designs using a Tailwind breakpoint


components/BackButton.tsx
```diff
@@ -1,14 +1,15 @@
 import { FC } from 'react';
 import { LeftChevron } from './icons';
-import { useRouter } from 'next/router';
 
-const BackButton: FC = () => {
-  const router = useRouter();
+interface Props {
+  callback: () => void;
```
AndrewLester: Rename for consistency with actual `button`. Make other occurences match.
```suggestion
  onClick: () => void;
```

components/Button.tsx
```diff
@@ -1,20 +1,20 @@
 import { FC, PropsWithChildren } from 'react';
 
 interface Props {
-  enabled: boolean;
+  disabled: boolean;
```
AndrewLester, line 4: Thanks for doing this

components/LocationInput.tsx
```diff
@@ -8,126 +8,164 @@ import React, {
 } from 'react';
 import mapboxgl from 'mapbox-gl';
 import MapboxGeocoder from '@mapbox/mapbox-gl-geocoder';
-
+import { LatitudeLongitude } from '~/lib/types';
 interface LocationInputProps {
```
AndrewLester, line 13: Formatting

```suggestion

interface LocationInputProps {
```

components/LocationInput.tsx
```diff
@@ -8,126 +8,164 @@ import React, {
...
+              offset: [0, 10],
+              anchor: 'top',
+              closeOnClick: false,
+              closeButton: false,
+              className: 'site-map',
+            })
+              .setLngLat([savedLng, savedLat])
+              .setHTML('<p>Drag map to pinpoint exact location.</p>')
+              .addTo(map.current);
+          }
```
AndrewLester, line 170: This looks great.

pages/site-details.tsx
```diff
@@ -0,0 +1,69 @@
...
+  Details = 'Details',
+  Location = 'Location',
+}
+
+interface Props {
+  page: Page;
+  setPage: (page: Page) => void;
+}
+
+const SiteDetailsDesktop: FC<Props> = ({ page, setPage }) => {
```
AndrewLester: This is the same as the function below named `SiteDetailsMobile`. I think you did too good of a job abstracting the logic, maybe remove one of these and rename the remaining one to `SiteForm` and use it in both occurances of the old ones.

lib/types.ts
```diff
@@ -43,3 +43,35 @@ export interface UnparsedForecastDataPoint {
...
+  longitude: number;
+}
+
+export interface Form {
+  siteCoordinates: LatitudeLongitude;
+  panelDetails: PanelDetails;
+  setFormData: ({ siteName, direction, tilt, capacity }: PanelDetails) => void;
+  setSiteCoordinates: ({ latitude, longitude }: LatitudeLongitude) => void;
+  postPanelData: () => void;
+}
```
AndrewLester, line 57: Formatting
```suggestion

}
```

components/form/Location.tsx
```diff
@@ -0,0 +1,69 @@
...
+            originalLng={siteCoordinates.longitude}
+            setIsSubmissionEnabled={setIsSubmissionEnabled}
+            setMapCoordinates={setSiteCoordinates}
+            zoomLevelThreshold={zoomLevelThreshold}
+            initialZoom={4}
+            canEdit={true}
+          />
+        </div>
+        <div className="md:hidden flex justify-center md:justify-end md:mt-20 mt-10 self-center w-full h-14">
+          <Button disabled={!isSubmissionEnabled} onClick={onClick}>
```
AndrewLester, line 55: Can be replaced with this like you have below? Then you can remove the `onClick` function definition in here.
```suggestion
          <Button disabled={!isSubmissionEnabled} onClick={nextPageCallback}>
```

components/form/Location.tsx
```diff
@@ -0,0 +1,69 @@
+import { useState, FC } from 'react';
+import Button from '~/components/Button';
+import LocationInput from '~/components/LocationInput';
+
+import { useFormContext } from '~/lib/context/form_context';
+import { originalLat, originalLng } from '~/lib/utils';
+import { zoomLevelThreshold } from '~/lib/utils';
+interface Props {
```
AndrewLester, line 8: Formatting
```suggestion

interface Props {
```

components/form/Location.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <Button disabled={!isSubmissionEnabled} onClick={onClick}>
+            Next
+          </Button>
+        </div>
+      </div>
+      <div className="hidden absolute md:flex md:flex-row md:justify-end bottom-12 left-1/2 -translate-x-1/2 w-3/4 h-14">
+        <Button disabled={!isSubmissionEnabled} onClick={nextPageCallback}>
+          Next
+        </Button>
+      </div>
```
AndrewLester, line 64: There's a bit of clipping with this button on my macbook air (not 1920x1080):
image

Perhaps instead of absolutely positioning this, you could just reuse the above button (line 55), and on `md` screens give it a width equal to a value _slightly_ bigger than the location input? Put this width on the div above it, then you should be good.

components/form/Details.tsx
```diff
@@ -0,0 +1,197 @@
...
+          </form>
+        </div>
+      </div>
+    </>
+  );
+};
+
+export default Details;
+
+export const getServerSideProps = withPageAuthRequired();
```
AndrewLester: Remnant from page component.
```suggestion

```

components/form/Details.tsx
```diff
@@ -0,0 +1,197 @@
...
+            <Modal show={showModal} setShow={setShowModal} />
+            <div className="hidden absolute md:flex md:flex-row md:justify-between bottom-12 left-1/2 -translate-x-1/2 w-3/4 h-14">
+              <Button disabled={false} onClick={lastPageCallback}>
+                Back
+              </Button>
+              <Button disabled={didSubmit}>
+                {didSubmit && <Spinner width={5} height={5} margin={2} />}
+                Finish
+                {didSubmit && <div className="w-5 mx-2" />}
+              </Button>
```
AndrewLester: Similar clipping on these buttons to on the Location page (with fixing steps mentioned below):
image

IMO remove the position `absolute` on the div above this, that should do most of the work. I just did some deeper looking tho and honestly I'd move this button row outside the form. That way, you can put it at the bottom of the page and give it full width and appropriately justify-between the buttons. You can tie the submit button here back to the form by giving the form an `id` and using the [form attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#:~:text=The%20%3Cform%3E%20element%20to%20associate%20the%20button%20with%20(its%20form%20owner).%20The%20value%20of%20this%20attribute%20must%20be%20the%20id%20of%20a%20%3Cform%3E%20in%20the%20same%20document.%20(If%20this%20attribute%20is%20not%20set%2C%20the%20%3Cbutton%3E%20is%20associated%20with%20its%20ancestor%20%3Cform%3E%20element%2C%20if%20any.))

components/LocationInput.tsx
```diff
@@ -8,126 +8,164 @@ import React, {
...
+    shouldZoomIntoOriginal,
+    originalLat,
+    originalLng,
     setIsSubmissionEnabled,
-    setLatExternal,
-    setLngExternal,
+    setMapCoordinates,
     zoom,
     zoomLevelThreshold,
   ]);
```
AndrewLester, line 185: A future issue: I think we should remove most if not all of these dependencies. The map ref should strictly interacted with via its set/get methods since recreating it might be causing issues... not for this PR.

AndrewLester: E.x. I'm noticing dragging the map doesn't exactly drag the marker the same way anymore... not too big a deal for now since it works well with lat/lng

AndrewLester (approved):
This is really really nice, thanks for grinding out the many style changes and mapbox interactions.

ericcccsliu (changes requested):
It seems that, once I enter a location, the location input automatically goes to the next step (is this location exact?). Could we wait until the user clicks "next" before going to the next step of pinpointing exact location?

lib/utils.ts
```diff
@@ -184,3 +150,50 @@ export const getNextThresholdIndex = (
     ),
   };
 };
+
+/* Represents the threshold for the graph */
+export const graphThreshold = 0.7;
+
+/* Latitude/longitude for London, England */
+export const originalLat = 51.5072;
+export const originalLng = 0.1276;
```
ericcccsliu, line 159: Is there a better place to define these? Don't really like hiding them away in a utils file. Perhaps define these in the component where they're used as defaults?

components/LocationInput.tsx
```diff
@@ -8,126 +8,164 @@ import React, {
...
         marker: false,
         reverseGeocode: true,
       });
-      if (geocoderContainer.current) {
-        geocoderContainer.current.appendChild(geocoder.onAdd(map.current));
+
+      if (!canEdit) {
+        // Prevent user from changing the geocoder input when the map isn't editable
+        geocoder.setBbox([0, 0, 0, 0]);
       }
```
ericcccsliu, line 80: nice touch

ashayp22:
> It seems that, once I enter a location, the location input automatically goes to the next step (is this location exact?). Could we wait until the user clicks "next" before going to the next step of pinpointing exact location?

This isn't intended behavior. Can you or @AndrewLester reproduce this bug?

components/LocationInput.tsx
```diff
@@ -8,126 +8,164 @@ import React, {
...
 
+      geocoderContainer.current?.appendChild(geocoder.onAdd(map.current));
```
ericcccsliu, line 82: According to the designs, we should be adding the geocoder on top of the map. However, I think this is fine for now—the position of the geocoder above the map might work better and look better in the long run. We do still need to style the geocoder according to the designs though. Could you try to do this using the styles in `globals.css`?

components/LocationInput.tsx
```diff
@@ -8,126 +8,164 @@ import React, {
...
+          if (popup === null && map.current && canEdit) {
+            popup = new mapboxgl.Popup({
+              offset: [0, 10],
+              anchor: 'top',
+              closeOnClick: false,
+              closeButton: false,
+              className: 'site-map',
+            })
+              .setLngLat([savedLng, savedLat])
+              .setHTML('<p>Drag map to pinpoint exact location.</p>')
```
ericcccsliu, line 168: Could we make this change to the "Drag map to pinpoint exact location" step be triggered by clicking the "next" button (which can perhaps be enabled after we're past the requisite zoom threshold) rather than having the transition be automatic?

ericcccsliu (changes requested):
Additionally, when I click the "back" button on the form details step, the map has to zoom/fly all the way back to the location I previously set. Could we make it so that upon pressing the "back" button, the map positions itself on the previous location by default, without any "flying" necessary? 

Screen Recording 2023-03-25 at 11 19 37 AM

components/form/Location.tsx
```diff
@@ -0,0 +1,69 @@
...
+    : 'Where is your solar panel located?';
+
+  return (
+    <>
+      <div
+        className="flex flex-col gap-2 relative h-[calc(100vh-var(--nav-height))] w-screen bg-mapbox-gray-1000"
+        id="rootDiv"
+      >
+        <div className="flex flex-col justify-end h-16 pl-3"></div>
+        <div
```
ericcccsliu, line 36: tiny thing but can we make the map have rounded edges

components/form/Details.tsx
```diff
@@ -0,0 +1,197 @@
...
+      <div className="flex flex-col">
+        <BackButton callback={lastPageCallback} />
+
+        <div className="flex flex-row w-full md:w-9/12 self-center">
+          <div className="flex-1 hidden md:block px-8">
+            {' '}
+            <h1 className="font-semibold text-4xl mt-2 dark:text-ocf-gray mb-5">
+              Your site&apos;s details
+            </h1>
+            <LocationInput
```
ericcccsliu: rounded edges would be nice here too

ericcccsliu:
> > It seems that, once I enter a location, the location input automatically goes to the next step (is this location exact?). Could we wait until the user clicks "next" before going to the next step of pinpointing exact location?
> 
> This isn't intended behavior. Can you or @AndrewLester reproduce this bug?

bug-in-location-input

lib/utils.ts
ashayp22: These constant values are being used in two different files, so we cannot define them in a single component as a default. We could export these values from that single component, but that is the same as what is going on here.

components/LocationInput.tsx
ashayp22: I don't understand why we'd need to do this from a user experience perspective. Wouldn't this be an unnecessary, unintuitive middle step? Can't we combine entering your address and making sure the pin is in the right spot into one step?

ashayp22: Why this could hurt the user experience:

- If the user needs to change their address, they'd need to go back 2 steps instead of 1.
- Most users won't need to adjust the pin since the solar panel will be on their house. Hence, we have an "optional" second step in the middle of the form.

ashayp22: Another concern: users may need to enter multiple addresses before confirming the location and moving the pin. However, how likely is this to happen? I argue that most homeowners will enter the correct address that first time, so automatically moving to the pinning step is fine.

ericcccsliu: > Why this could hurt the user experience:
> 
> * If the user needs to change their address, they'd need to go back 2 steps instead of 1.

I think this is pretty valid, but to address this, we could make it so that "back" on the metadata input form goes back to the first step, rather than the "pinpoint location" step in the middle

ericcccsliu: A bigger issue that might push us toward having that intermediary step rather than automatically moving forwards based on zoom value is that, if a user decides to input a location by manipulating the map, the transition to having a location pin show up is quite abrupt, and might be confusing, as the location pin appears when you're still quite far out from being able to pinpoint any exact location.

Even if this doesn't confuse most users, I believe the input behavior right now doesn't seem entirely well-polished or smooth. Furthermore, since this isn't an actual step, there is no "back" functionality—after the location pin appears, there's no way to revert back to the original state ("where is your solar panel located")

https://user-images.githubusercontent.com/62641231/227739333-ed45d377-e4c1-41d9-b904-bec16f62a4ec.mov

ericcccsliu: Also, to add another point—one of the main preliminary reasons that led to us even considering a redesign in the first place was to avoid having the "precise location" pin pop up based on threshold value in the first place.

ericcccsliu:
Could we disable the search box from showing suggestions with every change in reverse geocoding? Also, sometimes the pin on the map snaps to a location—could we disable that? 

https://user-images.githubusercontent.com/62641231/227740471-2452488f-551c-44dd-a735-bf6a97279fcf.mov

lib/utils.ts
ericcccsliu: hmm, might be better then to put these in a single component and export them from there? @AndrewLester what do you think? I just don't really like them being in a utils file.

ashayp22:
> Could we disable the search box from showing suggestions with every change in reverse geocoding? Also, sometimes the pin on the map snaps to a location—could we disable that?

I wasn't able to find a way to disable the search box from showing suggestions with every change in reverse geocoding. This should be made into its own issue. I also don't know how to resolve snapping to a location - it only happens with a few locations so we should investigate further in a separate issue.

ericcccsliu (approved):
👍

PR #132, opened 2023-03-21
Refactor Site page and SiteList Component to display SiteCards based on api data

### LINKED ISSUE #120: Incorporate site data into sites page
## Description
Our site page and site cards currently display hardcoded data. We need to have the site page display cards corresponding to all of a user's sites, and we need to have each card display relevant data about its respective site. 

Screen Shot 2023-03-16 at 1 44 00 PM

## Todos

- [x] populate site page with user's site list 
  - it might be beneficial to create a simple hook for the useSWR call for retrieving the site list
- [x] refactor cards to allow UUID to be passed in 
- [x] retrieve relevant data in cards using the `useSiteData` hook
- [x] display relevant data in cards 

## Useful links

- [figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=533%3A2251&t=4bIRt8hHMIjG4265-1)

## Time estimate
tbd

rohanvan123:
## Description
For this pull request, the main goal was to refactor the code to render the site card based on the list of UUIDs. The first portion of the code, I defined a method `ParseSiteUUIDs(data)` that takes in the list of siteUUIDs from the mock api call and converts it into a string array of exclusive the siteUUIDs. Then, that information is passed into a SiteCardLink component (which I changed the props for) to retrieve and display the correct data. I also added 3 more siteUUIDs (which are just duplicates of the first) to the site_uuids file. 

Fixes #120

### DIFF
--- a/components/SiteCard.tsx
+++ b/components/SiteCard.tsx
@@ -35,11 +35,13 @@ const SiteCard: FC<SiteCardProps> = ({ href, siteUUID }) => {
               Max. capacity: {installed_capacity_kw} kW
             </p>
           )}
-          {installed_capacity_kw && (
-            <p className="text-ocf-gray-500 font-medium text-xs">
-              Current yield: {currentOutput / installed_capacity_kw}%
-            </p>
-          )}
+          <p className="text-ocf-gray-500 font-medium text-xs">
+            Current yield:{' '}
+            {installed_capacity_kw
+              ? currentOutput / installed_capacity_kw
+              : currentOutput}
+            %
+          </p>
         </div>
       </div>
       {!href && (
--- a/components/SiteCard.tsx
+++ b/components/SiteCard.tsx
@@ -17,7 +17,7 @@ const SiteCard: FC<SiteCardProps> = ({ href, siteUUID }) => {
 
   const currentOutput = forecastData
     ? getCurrentTimeForecast(forecastData.forecast_values)
-    : 0;
+    : undefined;
 
   return (
     <a
@@ -28,7 +28,9 @@ const SiteCard: FC<SiteCardProps> = ({ href, siteUUID }) => {
         <h2 className="text-amber text-xl font-semibold">{client_site_name}</h2>
         <div className="flex flex-col mt-2 gap-1">
           <p className="text-ocf-gray-500 text-xs font-medium">
-            Current output: {currentOutput} kW
+            {`Current output: ${
+              currentOutput != undefined ? currentOutput + ' kW' : 'loading...'
+            }`}
           </p>
           {installed_capacity_kw && (
             <p className="text-ocf-gray-500 font-medium text-xs">
@@ -36,11 +38,12 @@ const SiteCard: FC<SiteCardProps> = ({ href, siteUUID }) => {
             </p>
           )}
           <p className="text-ocf-gray-500 font-medium text-xs">
-            Current yield:{' '}
-            {installed_capacity_kw
-              ? currentOutput / installed_capacity_kw
-              : currentOutput}
-            %
+            {`Current yield:
+            ${
+              installed_capacity_kw && currentOutput != undefined
+                ? currentOutput / installed_capacity_kw + '%'
+                : 'loading...'
+            }`}
           </p>
         </div>
       </div>
--- a/components/SiteCard.tsx
+++ b/components/SiteCard.tsx
@@ -1,53 +1,72 @@
 import { FC, PropsWithChildren } from 'react';
 import { DeleteIcon, EditIcon } from './icons';
 import Link from 'next/link';
+import { useSiteData } from '~/lib/hooks';
+import { getCurrentTimeForecast } from '~/lib/utils';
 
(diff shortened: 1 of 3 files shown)


components/SiteCard.tsx
```diff
@@ -1,53 +1,72 @@
...
-        <p className="text-ocf-gray-500 font-medium text-xs">
-          Max. capacity: 2800 kWh
-        </p>
+const SiteCard: FC<SiteCardProps> = ({ href, siteUUID }) => {
+  const { forecastData, client_site_name, installed_capacity_kw } =
+    useSiteData(siteUUID);
+
+  const currentOutput = forecastData
+    ? getCurrentTimeForecast(forecastData.forecast_values)
+    : 0;
```
ericcccsliu: instead of making this 0, can we make this `undefined`? having this be zero where it might not actually be zero seems a bit inaccurate.

components/SiteCard.tsx
```diff
@@ -1,53 +1,72 @@
...
+  return (
+    <a
+      href={href}
+      className="h-fit w-full max-w-lg flex bg-ocf-gray-1000 p-3 rounded-lg font-bold"
+    >
+      <div className="flex flex-col flex-1">
+        <h2 className="text-amber text-xl font-semibold">{client_site_name}</h2>
+        <div className="flex flex-col mt-2 gap-1">
+          <p className="text-ocf-gray-500 text-xs font-medium">
+            Current output: {currentOutput} kW
```
ericcccsliu: if we make the change in my comment above, this would have to show something like "loading..." if `currentOutput` is `undefined`

components/SiteCard.tsx
```diff
@@ -1,62 +1,74 @@
...
+          {installed_capacity_kw && (
+            <p className="text-ocf-gray-500 font-medium text-xs">
+              Max. capacity: {installed_capacity_kw} kW
+            </p>
+          )}
+          {installed_capacity_kw && (
+            <p className="text-ocf-gray-500 font-medium text-xs">
+              Current yield: {currentOutput / installed_capacity_kw}%
+            </p>
+          )}
```
AndrewLester: I'm now realizing what I told you about this earlier was wrong... my bad. Here's the process for showing current yield:
- If `installed_capacity_kw` is undefined, directly use `currentOutput` as "current yield"
- If it's defined, use the equation you have above.

pages/sites.tsx
```diff
@@ -3,10 +3,27 @@ import { EditIcon } from '~/components/icons';
 import { useState } from 'react';
 import SiteCardLink from '~/components/SiteCard';
 import { withSites } from '~/lib/utils';
+import { SiteList } from '~/lib/types';
+import useSWR from 'swr';
 
+const ParseSiteUUIDs = (data: SiteList): string[] => {
```
AndrewLester: Camel case, also change usage below.
```suggestion
const parseSiteUUIDs = (data: SiteList): string[] => {
```

pages/sites.tsx
```diff
@@ -3,10 +3,27 @@ import { EditIcon } from '~/components/icons';
...
+  const res = [];
```
AndrewLester: Rename this for better readability. Also rename usages below.
```suggestion
  const siteUUIDs = [];
```

AndrewLester (approved):
Really great work, a few small things.

pages/sites.tsx
```diff
@@ -19,10 +36,9 @@ const Sites = () => {
           )}
         </button>
       </div>
-      <SiteCardLink isEditMode={editMode} />
-      <SiteCardLink isEditMode={editMode} />
-      <SiteCardLink isEditMode={editMode} />
-      <SiteCardLink isEditMode={editMode} />
+      {siteUUIDs.map((site_uuid: string, idx: number) => (
+        <SiteCardLink key={idx} siteUUID={site_uuid} isEditMode={editMode} />
+      ))}
```
ericcccsliu: great catch to put `key={idx}` in, and this definitely works, but this can lead to things breaking if we aren't careful
 
the `key` property is meant to allow React to keep track of distinct objects. when using the index of each element in the list as a key, if the underlying list of `siteUUIDs` changes, React might easily lose track. You can find specific examples of why this might cause error [in this article](https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/). 

so instead, the `key` should be something more stable, and more unique to each element within `siteUUIDs`. a better option would naturally then be the `siteUUID` itself, as this would definitely be more permanent and unique. 

also—just another tiny detail, I think that any variables that we create ourselves should be camelcase, as that's what we're using throughout the entire codebase (other than some variable names that come from the API, which we can't change)

```suggestion
      {siteUUIDs.map((siteUUID: string) => (
        <SiteCardLink key={siteUUID} siteUUID={siteUUID} isEditMode={editMode} />
      ))}
```

ericcccsliu (approved):
great work—just a few small changes, and then you can go ahead and merge this in

PR #135, opened 2023-03-22
Update issue template to make Grammarly happy

ericcccsliu:
grammarly keeps flagging this as an error and I can't take it anymore


AndrewLester (approved):
Maybe we should update the main template too.

PR #139, opened 2023-03-24
Refactor sitecard components to link to unique dashboard pages with dynamic routing

### LINKED ISSUE #137: Display corresponding dashboard pages for each unique site card
## Description
For this feature, our goal is for a user to be able to click on a specific site card and be rerouted to its corresponding Dashboard page with the correct information. Currently, each site card links to the default dashboard page that we have set up. However, we want site cards to link specifically to the right dashboard based on the site's UUID. 

A straightforward approach is to let Next.js file/dynamic routing do most of the work for us, we want to have some type of identifiable key parameter in the link that our site card contains. An example of this would be "/dashboard/{site_uuid}", where using some Next.js library, we can store the site_uuid string and then render our dashboard page based off of it. This would also require modifying the href tag for site card components as they will no longer link to the default dashboard but the new route. 

## Todos

- [ ] Create some dashboard/[uuid] folder route such that [uuid] can be any string, assuming it is a valid uuid (don't worry about edge cases for now)
- [ ]  Retrieve that specific [uuid] value with the useRouter hook
- [ ] Render a dashboard based on the [uuid] value
- [ ] Alter site card components to link to our new dynamic routes

## Useful links
- [Next.js Dynamic Routes](https://nextjs.org/docs/routing/dynamic-routes)
- [Retrieving current path with Next](https://stackoverflow.com/questions/64836086/accessing-parameter-from-dynamic-route-in-next-js)

rohanvan123:
## Description
For this issue, the goal was for a user to be able to click on a site card link and be rerouted to its corresponding dashboard page. I first created a dashboard directory in the pages directory. The new folder contains a [uuid].tsx file and index.tsx file. The index returns the default dashboard "/dashboard", and the [uuid].tsx is the dynamic route file that considers the parameter [uuid]. I changed Dashboard to now have a siteUUID prop, this way we can generate based on the actual UUID itself. Finally, I changed the SiteCardLink component's href tag to point to /dashboard/[uuid]

Fixes #137


components/dashboard.tsx
```diff
@@ -1,17 +1,20 @@
...
+import { FC } from 'react';
 
-const siteUUID = 'b97f68cd-50e0-49bb-a850-108d4a9f7b7e';
-
-const Dashboard = () => {
+interface DashboardProps {
+  siteUUID: string;
+}
+const Dashboard: FC<DashboardProps> = ({ siteUUID }) => {
+  console.log(siteUUID);
```
ericcccsliu: remove this

pages/dashboard/index.tsx
```diff
@@ -0,0 +1,10 @@
+import Dashboard from '~/components/dashboard';
+import { withSites } from '~/lib/utils';
+
+const AggregateDashboard = () => {
+  const testUUID = 'b97f68cd-50e0-49bb-a850-108d4a9f7b7e';
+  return <Dashboard siteUUID={testUUID} />;
+};
```
ericcccsliu, line 9: can we add a comment saying that this is all temporary for now?

ericcccsliu (approved):
great work!

PR #148, opened 2023-03-25
Fix timezones and some fetching

AndrewLester:
## Status:

:rocket: Ready

## Description

Fix some timezone issues. Before this, nowcasting forecast times were being parsed in the user's local timezone. Since the sites we're working with _are_ actually in UTC, this effectively made the graphs display the forecasts in the site timezone (that's funny). Parsing in the user's local timezone won't work for any site outside UTC though, so now I've changed the code to make sure we're parsing the times in UTC, which _correctly_ makes the graph show the forecasts in UTC. We can in the future change the graph's x-axis to show correct timezone using this: https://stackoverflow.com/a/62155199.

### DIFF
--- a/components/graphs/Graph.tsx
+++ b/components/graphs/Graph.tsx
@@ -12,6 +12,7 @@ import { formatter, forecastDataOverDateRange } from 'lib/graphs';
 import { useSiteData } from 'lib/hooks';
 import useTime from '~/lib/hooks/useTime';
 import { FC } from 'react';
+import { ForecastDataPoint } from '~/lib/types';
 
 const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
   const { forecastData, latitude, longitude } = useSiteData(siteUUID);
@@ -19,13 +20,13 @@ const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
 
   const endDate = new Date();
   endDate.setHours(endDate.getHours() + 48);
-  const graphData = forecastData
-    ? forecastDataOverDateRange(
-        forecastData.forecast_values,
-        new Date(currentTime),
-        endDate
-      )
-    : [];
+  const graphData =
+    forecastData &&
+    forecastDataOverDateRange(
+      forecastData.forecast_values,
+      new Date(currentTime),
+      endDate
+    );
   const maxGeneration = graphData
     ? Math.max(...graphData.map((value) => value.expected_generation_kw))
     : 0;
@@ -62,8 +63,8 @@ const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
             dataKey="target_datetime_utc"
             stroke="white"
             axisLine={false}
-            tickFormatter={(point: string) =>
-              formatter.format(Date.parse(point))
+            tickFormatter={(point: ForecastDataPoint['target_datetime_utc']) =>
+              formatter.format(new Date(point))
             }
           />
           <YAxis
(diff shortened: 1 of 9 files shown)


pages/sites.tsx
```diff
@@ -24,11 +24,14 @@ const parseSiteUUIDs = (data: SiteList): string[] => {
...
   const [editMode, setEditMode] = useState(false);
 
-  const { data, error, isLoading } = useSWR(
-    `${process.env.NEXT_PUBLIC_API_BASE_URL}/api/sites`
+  const { data } = useSWR<SiteList>(
+    `${process.env.NEXT_PUBLIC_API_BASE_URL_GET}/sites`
   );
 
-  const siteUUIDs: string[] = parseSiteUUIDs(data);
+  // TODO: Paginate this or something... when pulling from actual API it's just too many
```
ericcccsliu, line 31: Could you open an issue and link it in this TODO? Doesn't have to be written out, just would like it to be up on our board so we won't forget about this.

pages/sites.tsx
ericcccsliu, line 1: Screen Shot 2023-03-25 at 11 32 02 AM
perhaps we could add a default value for site name so our site cards don't look too sad here

ericcccsliu, line 1: Screen Recording 2023-03-25 at 11 37 28 AM
Might need better loading functionality here—could be opened up as a new issue

components/graphs/Graph.tsx
```diff
@@ -74,18 +75,20 @@ const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
             labelStyle={{ color: 'white' }}
-            formatter={(value: number, name, props) => [
+            formatter={(value: ForecastDataPoint['expected_generation_kw']) => [
               parseFloat(value.toFixed(5)),
               'kW',
             ]}
-            labelFormatter={(point: string) =>
-              formatter.format(Date.parse(point))
+            labelFormatter={(point: ForecastDataPoint['target_datetime_utc']) =>
+              formatter.format(new Date(point))
```
ericcccsliu, line 91: woot woot

components/graphs/ThresholdGraph.tsx
```diff
@@ -269,11 +217,10 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
               stroke="white"
               strokeDasharray="2"
               fill="url(#colorUv)"
+              // Necessary for label list https://stackoverflow.com/a/55724641
+              // Better solution for this would be nice
+              isAnimationActive={false}
```
ericcccsliu, line 222: interesting, thx for digging this up. definitely makes things look less pretty but at least it works!

ericcccsliu (approved):

components/graphs/ThresholdGraph.tsx
AndrewLester: I'll make an issue for this too. It does look quite nice when it works

pages/sites.tsx
AndrewLester: https://github.com/openclimatefix/pv-sites-mobile/issues/152

pages/sites.tsx
AndrewLester: I see you have one: https://github.com/openclimatefix/pv-sites-mobile/issues/151

components/graphs/ThresholdGraph.tsx
AndrewLester: https://github.com/openclimatefix/pv-sites-mobile/issues/153

PR #159, opened 2023-03-26
Attempt graph animation fix and add 'now indicator'

### LINKED ISSUE #153: Figure out why the `LabelList` that shows current time dot doesn't show with graph animations enabled
**Describe the bug**

The `LabelList` that shows the current time as a yellow dot (on the `ThresholdGraph`) doesn't show up when graph animations are enabled. Here's a relevant [stackoverflow thread](https://stackoverflow.com/a/55724641).

**To Reproduce**

1. Go to the main branch
2. Remove the [isAnimationEnabled={false} line](https://github.com/openclimatefix/pv-sites-mobile/blob/2c875c4d3a40eaca4dab4b895bfa298e055868a9/components/graphs/ThresholdGraph.tsx#L220-L222)
3. Note that the dot doesn't show

**Expected behavior**

It shows up, with animations

We need to dig into this more and figure out how we can get them both working in an _elegant_ fashion.

AndrewLester:
## Status:

🚀 Ready

## Description

Fix animation by disabling `useTime` while graph data loads. Add now indicator to lower graph

Fixes: #153

## Screenshots

image

### DIFF
--- a/components/graphs/Graph.tsx
+++ b/components/graphs/Graph.tsx
@@ -34,7 +34,7 @@ const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
     useSiteData(siteUUID);
   const [timeEnabled, setTimeEnabled] = useState(false);
   const { currentTime } = useTime(latitude, longitude, {
-    enabled: timeEnabled,
+    updateEnabled: timeEnabled,
   });
 
   const endDate = new Date();
--- a/components/graphs/Graph.tsx
+++ b/components/graphs/Graph.tsx
@@ -1,22 +1,72 @@
 import {
   CartesianGrid,
+  Label,
   Line,
   LineChart,
+  ReferenceLine,
   ResponsiveContainer,
   Tooltip,
   XAxis,
   YAxis,
 } from 'recharts';
 import { LegendLineGraphIcon } from '@openclimatefix/nowcasting-ui.icons.icons';
-import { formatter, forecastDataOverDateRange } from 'lib/graphs';
+import {
+  formatter,
+  forecastDataOverDateRange,
+  getCurrentTimeForecastIndex,
+} from 'lib/graphs';
 import { useSiteData } from 'lib/hooks';
 import useTime from '~/lib/hooks/useTime';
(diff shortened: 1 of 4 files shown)


components/graphs/Graph.tsx
```diff
@@ -39,67 +58,101 @@ const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
+    };
+    return (
+      <g>
+        <g className="fill-white">
+          <text stroke="white" strokeWidth="1em" {...textProps}>
+            {/* Use invisible "H" character with stroke width to create white background around text */}
+            {'H'.repeat('Now'.length)}
+          </text>
+          <text {...textProps}>Now</text>
+        </g>
```
AndrewLester, line 77: Most changes in this PR are just to make animations work and are small (is `isLoading` before rendering graph). This is a weirder one, to add a background color of white on the text I had to create another text element and give that element a stroke width, that appears around the characters. Since the 'H' has straight lines, so does the stroke. Apparently SVG's don't do `background-color`.

lib/hooks/useTime.ts
```diff
@@ -22,13 +26,15 @@ const useTime = (latitude?: number, longitude?: number) => {
...
-  });
+    if (enabled) {
+      const intervalId = setInterval(() => {
+        setCurrentTime(Date.now());
+      }, 1000);
+
+      // clear interval on re-render to avoid memory leaks
+      return () => clearInterval(intervalId);
+    }
+  }, [enabled]);
```
AndrewLester: Another thing to note is the implementation of `enabled` for `useTime`. `useTime`'s interval can now be disabled to prevent rerendering.

ericcccsliu: could we use a more descriptive name? perhaps something like `updateEnabled` might suffice.

components/graphs/Graph.tsx
ericcccsliu: oh weird

ericcccsliu (approved):

ericcccsliu:
looks good locally!!

PR #164, opened 2023-03-27
Implement differing time ranges on graph

### LINKED ISSUE #142: Implement differing time ranges on graph
## Description
Add buttons to our graph to set the time range displayed. Buttons should be a radio group, and time range should be restricted accordingly. 

Screen Shot 2023-03-23 at 8 57 03 PM

Some additional context about the time ranges. The "now indicator", that is/will be added in #159, sits around 1/8 the way into the graph from the left side. We should keep it this way (or switch to another fraction if it looks better), while respecting the time range selection. You'll have to perform a calculation to get the range start and end times based on the the time range _and the current time_.

## Todos

- [x] Create radio buttons 
- [x] Restrict time range of graph based on button selected 

## Useful links

- [figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=1096%3A3089&t=SG5p3vUB6TcI21bD-1)

## Time estimate
tbd!

neha-vard:
## Description
Add buttons to our graph to set the time range displayed. Buttons should be a radio group, and time range should be restricted accordingly.

Fixes #142 

## Screenshots
Screen Shot 2023-04-03 at 1 02 11 PM
Screen Shot 2023-04-03 at 1 02 33 PM


components/graphs/Graph.tsx
```diff
@@ -114,7 +135,51 @@ const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
+          <input
+            className="hidden peer"
+            type="radio"
+            name="1H"
+            id="1H"
+            value="1"
+            checked={timeRange === '1'}
+            onChange={handleChange}
+          />
+          <span className="peer-checked:bg-ocf-yellow-500 peer-checked:rounded-md peer-checked:text-black text-ocf-gray-300 w-10 h-7 pt-0.5 text-center bg-ocf-gray-1000 rounded-md inline-block relative">
```
AndrewLester: Maybe give `cursor-pointer` to all these spans?

components/graphs/Graph.tsx
```diff
@@ -56,21 +72,26 @@ const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
     updateEnabled: timeEnabled,
   });
 
+  const [timeRange, setTimeRange] = useState('24');
```
AndrewLester: Replace this string with a number so you don't have to do `parseInt` down below. There's almost no difference between using a number or a string here, doing it this way will just reduce the number of `parseInt` calls.
```suggestion
  const [timeRange, setTimeRange] = useState(24);
```

components/graphs/Graph.tsx
```diff
@@ -56,21 +72,26 @@ const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
+  const handleChange = (event: any) => {
+    setTimeRange(event.target.value);
```
AndrewLester, line 77: Do the parseInt here instead
```suggestion
    setTimeRange(parseInt(event.target.value));
```

components/graphs/Graph.tsx
```diff
@@ -114,7 +135,51 @@ const Graph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
+    <div className="w-full h-[290px] bg-ocf-black-500 rounded-2xl p-1">
+      <div className="mt-3 ml-1">
+        <label className="mx-2">
+          <input
+            className="hidden peer"
+            type="radio"
+            name="1H"
+            id="1H"
+            value="1"
+            checked={timeRange === '1'}
```
AndrewLester: Make sure to replace these comparisons with the number `1` (and others below), after changing the state variable to store a number.

AndrewLester (approved):
Great work, the computations for this must have been annoying. Just one very small change with the useState type.

components/graphs/Graph.tsx
```diff
@@ -19,13 +19,29 @@ import {
...
   const currentDate = new Date(currentTime);
   return new Date(
     currentDate.getFullYear(),
     currentDate.getMonth(),
     currentDate.getDate(),
-    currentDate.getHours() - 5
+    totalHours > 1
+      ? currentDate.getHours() - totalHours / 8
+      : currentDate.getHours(),
+    totalHours > 1 ? 0 : currentDate.getMinutes() - 15
```
ericcccsliu: Could we perhaps get a comment above this function describing what this bit of math does?

components/graphs/Graph.tsx
```diff
@@ -19,13 +19,29 @@ import {
...
+function getGraphEndDate(currentTime: number, totalHours: number) {
+  const currentDate = new Date(currentTime);
+  return new Date(
+    currentDate.getFullYear(),
+    currentDate.getMonth(),
+    currentDate.getDate(),
+    totalHours > 1
+      ? currentDate.getHours() + (7 * totalHours) / 8
+      : currentDate.getHours(),
+    totalHours > 1 ? 0 : currentDate.getMinutes() + 45
```
ericcccsliu: same here, everything's working great but would like description of behaviour in a comment for maintainability

ericcccsliu (approved):
absolutely :fire: work. we might need some comments to describe the non-straightforward calculations we're doing, but otherwise everything looks and works great!

ericcccsliu (approved):
Great work! Still one tiny comment from Andrew that seems like it needs to be addressed, and for future reference, make sure to include a couple of screenshots of your great work, so we can initially verify it at a glance.

ericcccsliu:
> Great work! Still one tiny comment from Andrew that seems like it needs to be addressed, and for future reference, make sure to include a couple of screenshots of your great work, so we can initially verify it at a glance.

actually—ignore the comment for now haha

PR #171, opened 2023-03-29
Fix withSites

AndrewLester:
## Status:

:rocket: Ready

## Description

This might fix the Auth0 error you sometimes see when using the site. I'm not sure why this is necessary... sometimes the access token needs to be refreshed, but the user was able to pass through `withPageAuthRequired`? This might not even fix it. Either way, it fixes another issue (unawaited promise)... doesn't have an effect right now since our mock api doesn't deal with the token anyway.


ericcccsliu (approved):
LGTM

Issue #160, opened 2023-03-27
Tie threshold graph start and end times to sun event times calculated by Suncalc


AndrewLester:
## Description

The threshold graph's start and end times are currently fixed at 8am and 8pm respectively.
https://github.com/openclimatefix/pv-sites-mobile/blob/afd49547091510610b282d6fb29de62e79ecff34/lib/graphs.ts#L52-L74

These times don't _always_ correspond to starting at 0 solar output and ending at 0, since the season changes the sunrise/sunset times. We should set the start and end time based on the sun rise and sunset time calculated in `useTime`. 
https://github.com/openclimatefix/pv-sites-mobile/blob/afd49547091510610b282d6fb29de62e79ecff34/lib/hooks/useTime.ts#L21-L22

You'll have to expose these times from the hook, and use them to filter the forecast data. Also leave 30 mins of buffer range before and after each event to leave room for output to hit 0. Maybe also round everything to the nearest 30 min interval.

## Todos

These points are rough guidelines. **Please** feel free to discuss the best way to design this component with others on the team!

- [ ] Expose the sunrise and sunset times from the hook
- [ ] Leave 30 mins of buffer before the sunrise and 30 mins after the sunset.
- [ ] Round these times off to the nearest 30 min increment that still fits the criterion above
- [ ] Use these times to filter the forecast data ([getClosestForecastIndex](https://github.com/openclimatefix/pv-sites-mobile/blob/afd49547091510610b282d6fb29de62e79ecff34/lib/graphs.ts#L17) may be useful)

## Time estimate

tbd

PR #180, opened 2023-03-31
Create a hook to format data based on the user's local time zone

### LINKED ISSUE #147: Show times in site's local timezone
## Description

Currently, times are displayed in the user's device timezone, but the site that the user is viewing might be in a separate timezone. In this case, it would make more sense to display times in the site's timezone.

We can use Mapbox for this to maintain library consistency throughout the app. Follow this tutorial to use their Tilequery API to get the site's timezone from its latitude and longitude: https://docs.mapbox.com/help/tutorials/create-a-timezone-finder-with-mapbox-tilequery-api/

## Todos

Use the Mapbox [Tilequery API](https://docs.mapbox.com/help/tutorials/create-a-timezone-finder-with-mapbox-tilequery-api/) in a hook that takes in a latitude and longitude (or just site data), to then return a timezone string `TZID` from Mapbox.

- [x] Create the hook that calls Mapbox's API
- [x] Update our [date formatter](https://github.com/openclimatefix/pv-sites-mobile/blob/2c875c4d3a40eaca4dab4b895bfa298e055868a9/lib/graphs.ts#L4) to take into account the timezone via a change like this [quick guide on stackoverflow](https://stackoverflow.com/questions/62154757/what-exactly-does-the-timezone-option-do-in-intl-datetimeformat/62155199#62155199)[^1]

## Useful links

- [Tilequery API](https://docs.mapbox.com/help/tutorials/create-a-timezone-finder-with-mapbox-tilequery-api/)

## Time estimate

[^1]: it might be best to pass just the timezone into the formatter to keep it clean, and retrieve the timezone using the hook that calls Mapbox's API wherever we need to use the formatters

rohanvan123:
## Description

For this issue, the general problem was that a time increments for both graph components were based on some hard coded timezone and not the actual location of the user site. I implemented a `useDataFormatter` hook that returns the timezone, timeFormatter function, and weekdayFormatter function. These functions take in a timezone parameter. The way this parameter is obtained is through a call of Mapbox's API. This call is wrapped around a use effect which has that long and lat as dependencies. 

Fixes #147 

## Todos
- This brings up another issue, which is that the times for the threshold graph are not from 8:00 to 8:00 PM locally (they adjust based on the timezone)

### DIFF
--- a/lib/graphs.ts
+++ b/lib/graphs.ts
@@ -1,16 +1,16 @@
-/**
- * Converts Date object into Hour-Minute format based on device region
- */
-export const timeFormatter = new Intl.DateTimeFormat(['en-US', 'en-GB'], {
-  hour: 'numeric',
-  minute: 'numeric',
-});
+// /**
+//  * Converts Date object into Hour-Minute format based on device region
+//  */
+// export const timeFormatter = new Intl.DateTimeFormat(['en-US', 'en-GB'], {
+//   hour: 'numeric',
+//   minute: 'numeric',
+// });
 
-export const weekdayFormatter = new Intl.DateTimeFormat(['en-US', 'en-GB'], {
-  weekday: 'short',
-  hour: 'numeric',
-  minute: 'numeric',
-});
+// export const weekdayFormatter = new Intl.DateTimeFormat(['en-US', 'en-GB'], {
+//   weekday: 'short',
+//   hour: 'numeric',
+//   minute: 'numeric',
+// });
 
 interface ForecastDataPoint {
   target_datetime_utc: number;
--- a/lib/graphs.ts
+++ b/lib/graphs.ts
@@ -9,12 +9,6 @@ export const timeFormatter = new Intl.DateTimeFormat(['en-US', 'en-GB'], {
   minute: 'numeric',
 });
 
-// export const weekdayFormatter = new Intl.DateTimeFormat(['en-US', 'en-GB'], {
-//   weekday: 'short',
-//   hour: 'numeric',
-//   minute: 'numeric',
-// });
-
 interface ForecastDataPoint {
   target_datetime_utc: number;
   expected_generation_kw: number;
--- a/lib/graphs.ts
+++ b/lib/graphs.ts
@@ -1,4 +1,3 @@
-
 import { ClearSkyDataPoint } from './types';
 
 /**
--- a/lib/hooks/useDateFormatter.ts
+++ b/lib/hooks/useDateFormatter.ts
@@ -2,7 +2,7 @@ import { useEffect, useState } from 'react';
 import useSiteData from './useSiteData';
 
 /**
- * Gets forecasted and solar panel data for a single site
+ * Gets the current timezone code based on a user's site location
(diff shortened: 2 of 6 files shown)


AndrewLester (approved):
Very nice job

ericcccsliu:
todos should be addressed and fixed by #160

lib/graphs.ts
```diff
@@ -8,11 +9,11 @@ export const timeFormatter = new Intl.DateTimeFormat(['en-US', 'en-GB'], {
...
-export const weekdayFormatter = new Intl.DateTimeFormat(['en-US', 'en-GB'], {
-  weekday: 'short',
-  hour: 'numeric',
-  minute: 'numeric',
-});
+// export const weekdayFormatter = new Intl.DateTimeFormat(['en-US', 'en-GB'], {
+//   weekday: 'short',
+//   hour: 'numeric',
+//   minute: 'numeric',
+// });
```
ericcccsliu: delete this comment

lib/hooks/useDateFormatter.ts
```diff
@@ -0,0 +1,51 @@
+import { useEffect, useState } from 'react';
+import useSiteData from './useSiteData';
+
+/**
+ * Gets forecasted and solar panel data for a single site
```
ericcccsliu: great having this comment here, but this line is inaccurate (copied/pasted but not changed perhaps?) and should be fixed

lib/hooks/useDateFormatter.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+const fetchTimeZone = async (long: number, lat: number) => {
+  const query = await fetch(
+    `https://api.mapbox.com/v4/examples.4ze9z6tv/tilequery/${long},${lat}.json?access_token=${process.env.NEXT_PUBLIC_MAPBOX_PUBLIC}`,
+    { method: 'GET' }
+  );
+  const data = await query.json();
+  // get the timezone from the resulting GeoJSON FeatureCollection
+  const userTimezone = data?.features[0]?.properties?.TZID ?? 'GB';
```
ericcccsliu, line 19: perhaps leave a comment here describing behavior when the time zone is not found—the assumption of `'GB'` might not be valid in the very long term, so we want to make clear that this is an assumption here.

lib/hooks/useDateFormatter.ts
```diff
@@ -0,0 +1,51 @@
...
+    weekday: 'short',
+    hour: 'numeric',
+    minute: 'numeric',
+    timeZone: timezone,
+  });
+
+  return { timezone, timeFormatter, weekdayFormatter };
+};
+
+export default useDateFormatter;
```
ericcccsliu, line 52: absolutely great work here

ericcccsliu (changes requested):
Looks great, great work. But, since there are still build errors happening, I can't go in and verify your changes on deployment or locally—could you fix those, address the (few) comments, and then re-request me so I can give a visual check?

ericcccsliu (approved):
lets get this merged

PR #181, opened 2023-03-31
Fix redirect

AndrewLester:
## Status:

:rocket: Ready

## Description

Fix redirect in withSites to use correct redirect value.


ericcccsliu (approved):
LGTM

PR #183, opened 2023-03-31
Fix bugs in site details

### LINKED ISSUE #178: Map input occasionally incorrectly disabling
**Describe the bug**
Next button occasionally greyed out/disabled, even when in the UK and map is not moving 
Screen Shot 2023-03-30 at 5 08 58 PM

**To Reproduce**

1. Go to a place in the UK
2. Zoom in far enough so that our pin appears
3. Move around
4. Eventually see error

**Expected behavior**
Whenever your pin is anywhere in the UK, and the map is not moving, the next button should be enabled.

### LINKED ISSUE #182: Sites card links not responding to click events
**Describe the bug**
On the sites card, the links aren't routing to their respective pages

ericcccsliu:
## Status:
:rocket: Ready

## Description
Fixes click events not occurring on the edit/delete buttons in the site card. This was accomplished by adding `passHref` to the `Link` components, so that the correct `href` would be passed down to the wrapped `a` tags, and adding `pointer-events-auto` conditionally to the buttons so that clickevents wouldn't be disabled. 

Also fixes our map input occasionally disabling by adding state to track if the map is moving or not, and removing setters from the `onMoveStart` listener, which seems to only trigger when a move is started under Mapbox control. 

Also changes name of `Sidebar` to `SideBar`, refactors components into folders, and removes some unused imports that were missed previously.

Fixes #178, 
fixes #182 

## Todos

We've got a console error since we have `a` tags wrapping `a` tags in the SiteCard. We need to open up a new issue to fix this—a draft is up on our board—as this fix would be nontrivial.


components/LocationInput.tsx
```diff
@@ -32,6 +32,7 @@ const LocationInput: FC<LocationInputProps> = ({
   const [isPastZoomThreshold, setIsPastZoomThreshold] =
     useState<boolean>(false);
   const [isInUK, setIsInUK] = useState<boolean>(false);
+  const [isMoving, setIsMoving] = useState<boolean>(false);
```
amit-sawhney: NIT: This might be a bit of overtyping. For cleanliness, let typescript infer some of these variable types. For example, you couldn't say (or I hope you wouldn't): 
```ts
const x : number = 2;
```
It's valid but bloats code a little bit potentially.

components/LocationInput.tsx
```diff
@@ -225,8 +224,8 @@ const LocationInput: FC<LocationInputProps> = ({
   ]);
 
   useEffect(() => {
-    setIsSubmissionEnabled(isInUK && isPastZoomThreshold);
-  }, [isInUK, isPastZoomThreshold, setIsSubmissionEnabled]);
+    setIsSubmissionEnabled(isInUK && isPastZoomThreshold && !isMoving);
+  }, [isInUK, isPastZoomThreshold, setIsSubmissionEnabled, isMoving]);
```
amit-sawhney, line 227: Is there a reason we are prop drilling setState functions? I feel like a callback design pattern makes more sense and is much more extensible. Let the parent handle the state updates in it's own file and simply pass in a call back that passes back the information that your parent needs to update the state.

components/dashboard-modules/SunnyTimeframe.tsx
```diff
@@ -12,7 +12,7 @@ const SunnyTimeframe: FC<{ siteUUID: string }> = ({ siteUUID }) => {
   const { forecastData } = useSiteData(siteUUID);
 
   if (!forecastData) {
-    return null;
+    return <NumberDisplay title={'Loading'} value={'...'} />;
```
amit-sawhney: NIT: Don't need to wrap string constants in brackets.

components/navigation/SideBar.tsx
```diff
@@ -45,34 +45,34 @@ const MenuLink: React.FC<MenuLinkProps> = ({
...
 
   return (
     <div
       className={`z-50 transition-all duration-500 h-full fixed top-0 ${
-        isSidebarOpen
+        isSideBarOpen
           ? 'translate-x-0 shadow-lg shadow-ocf-black'
           : '-translate-x-64'
       }`}
       // @ts-ignore
```
amit-sawhney, line 69: 🤨

lib/context/sidebar_context.tsx
```diff
@@ -1,38 +1,38 @@
...
-  closeSidebar: () => void;
+interface SideBar {
+  isSideBarOpen: boolean;
+  openSideBar: () => void;
+  closeSideBar: () => void;
 }
 
-interface SidebarProviderProps {
+interface SideBarProviderProps {
   children: ReactNode | ReactNode[];
```
amit-sawhney: `PropsWithChildrenType` @AndrewLester

pages/api/[...mockApiRoute].ts
```diff
@@ -1,4 +1,4 @@
-import { addMilliseconds, getDate } from 'date-fns';
+import { addMilliseconds } from 'date-fns';
```
Comment (amit-sawhney) [inline pages/api/[...mockApiRoute].ts:1]: good choice not to use `moment` lol

amit-sawhney (commented):
Can't comment too much on product specific functionality, but some code quality and style nits

AndrewLester (approved):

pages/api/[...mockApiRoute].ts
Comment (AndrewLester) [reply pages/api/[...mockApiRoute].ts:1]: Ikr, unfortunately this library has its own issues and I'm contemplating removing it (this took me a while to figure out): https://github.com/date-fns/date-fns/issues/3129

PR #184, opened 2023-04-01
Tie threshold graph to Suncalc

### LINKED ISSUE #160: Tie threshold graph start and end times to sun event times calculated by Suncalc

ashayp22:
## Description

Sets the start and end time of the threshold graph to the dusk/dawn times returned by Suncalc.

Fixes #160

### DIFF
--- a/lib/hooks/useTime.ts
+++ b/lib/hooks/useTime.ts
@@ -48,7 +48,6 @@ const useTime = (
     [currentTime, times]
   );
 
-  // Updates the
   useEffect(() => {
     if (updateEnabled) {
       const intervalId = setInterval(() => {
--- a/lib/hooks/useTime.ts
+++ b/lib/hooks/useTime.ts
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useMemo } from 'react';
 const SunCalc = require('suncalc');
 
 type UseTimeOptions = {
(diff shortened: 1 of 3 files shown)


AndrewLester (approved):
Nice work on this change 

One thing: build is failing currently since on Energy Recommendation.tsx, it uses `isDaytime`, but in yours is returns `isDayTime` right? Or maybe vice versa?

lib/hooks/useTime.ts
```diff
@@ -22,14 +24,31 @@ const useTime = (
...
+  const isDayTime = useMemo(
+    () =>
+      times !== null &&
+      currentTime >= times.dawnTime.getTime() &&
+      currentTime <= times.duskTime.getTime(),
+
+    [currentTime, times]
+  );
+
+  // Updates the
```
ericcccsliu: updates the what?

ericcccsliu (approved):
looks awesome, ty!!

PR #185, opened 2023-04-01
Display threshold graph current time marker between dawn and dusk

### LINKED ISSUE #165: Current time marker still rendered on ThresholdGraph when displaying next day's data (after 8 pm)
**Describe the bug**
The current time marker is still rendered on ThresholdGraph when displaying next day's data (after 8 pm)

**To Reproduce**
Look at ThresholdGraph in dashboard after 8 pm.

**Expected behavior**
Don't render the current time marker when it is past 8 pm.

**Additional context**
Proposed fix: inside the renderCurrentTimeMarker function, return null if it is past 8 pm. Can use same logic as getGraphStartDate and getGraphEndDate to determine if it is past 8 pm.

ashayp22:
## Description

Only display the Threshold graph's current time marker between the start and end times of the graph, which is the dawn/dusk times of wherever the solar site is located.

Fixes #165

### DIFF
--- a/components/graphs/ThresholdGraph.tsx
+++ b/components/graphs/ThresholdGraph.tsx
@@ -57,10 +57,17 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
   const renderCurrentTimeMarker = ({ x, y, index }: any) => {
     if (!graphData) return null;
 
+    /* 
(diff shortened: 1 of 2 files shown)


ericcccsliu (approved):
LGTM

AndrewLester (approved):

PR #186, opened 2023-04-02
General fixes

AndrewLester:
## Status:

:rocket: Ready

## Description

General fixes.
Deployment is now good.

## Notes
- One bug still, if you navigate to /sites then back to /dashboard, the current time indicator in the thresholdgraph doesn't show.


components/SiteCard.tsx
```diff
@@ -15,83 +14,68 @@ interface SiteCardProps {
...
+              <SiteGraph siteUUID={siteUUID} hidden={isEditMode} />
+            </div>
+          </div>
+        </a>
         <div
-          className={`transition-all ${
-            isEditMode ? 'w-4/12' : 'w-0'
-          } duration-[${transitionDuration}ms] flex translate-x-40`}
+          // @ts-ignore
+          inert={!isEditMode ? '' : null}
```
ericcccsliu, line 71: think the `@ts-ignore` is fine for now—tried to fix it but couldn't get anything. however, isn't `inert` a boolean attribute? 

```suggestion
          // @ts-ignore
          inert={!isEditMode ? false : true}
```
Don't know for sure if that'd be correct behavior.

components/dashboard-modules/NumberDisplay.tsx
```diff
@@ -3,16 +3,20 @@ import { FC } from 'react';
 interface Props {
   title: string;
   value: string;
+  onClick?: () => void;
 }
 
-const NumberDisplay: FC<Props> = ({ title, value }) => {
+const NumberDisplay: FC<Props> = ({ title, value, onClick }) => {
+  const Element = onClick ? 'button' : 'div';
```
ericcccsliu, line 10: nice addition

components/SiteCard.tsx
AndrewLester: I think tried just dropping the boolean in itself and ran into problems. This github comment seemed to simply this was the best way, though I could do '' : false. https://github.com/WICG/inert/issues/58#issuecomment-618016847

components/icons/future_threshold.tsx
```diff
@@ -1,34 +1,18 @@
 import { FC } from 'react';
 
 export const FutureThresholdLegendIcon: FC = ({}) => (
```
ericcccsliu: Maybe change the name of this and move it out of icons, since it isn't really an icon anymore? Not strictly necessary, feel free to disregard if you think otherwise.

lib/graphs.ts
```diff
@@ -1,89 +1,126 @@
...
+
+export function addTimePoint(
+  generationData: GenerationDataPoint[],
+  date: Date
+) {
+  const generationDataInterpolated = generationData.map((data) => ({
+    ...data,
+  }));
+
+  const forecastValuePeriod = 15;
```
ericcccsliu: Can we infer this from the data? Just need to take the difference in time between the closestTime and the next index. Would help when we implement actuals, which have more frequent data.

pages/dashboard/index.tsx
```diff
@@ -9,4 +9,16 @@ const AggregateDashboard = () => {
...
+      return {
+        redirect: {
+          permanent: false,
+          destination: `/dashboard/${siteList.site_list[0].site_uuid}`,
+        },
+      };
+    }
+    return { props: { siteList } };
+  },
+});
```
ericcccsliu, line 24: Think a better solution might be to display the `Dashboard` component with the `site_uuid` set to the ID of the only site—wouldn't need to redirect and would keep the routes simple for just one site.

ericcccsliu (approved):

components/icons/future_threshold.tsx
AndrewLester: Inlined it

pages/dashboard/index.tsx
AndrewLester: The reason I did this was because the view we'll have for a site dashboard vs. aggregate dashboard might very different. We could either load a different component, or more likely just pass a prop to the component. In that case, we really want to just treat someone who has one site's "aggregate dashboard" as their normal dashboard, and we can get that by just showing them the single site dashboard page. Then, linking to `/dashboard` in the sidebar works if we redirect them. Just some abstractions that I think make sense, lmk your thoughts.

AndrewLester:
Merging for now, we can open up issues for any additional comments.

PR #188, opened 2023-04-03
Skeleton loading screens

### LINKED ISSUE #151: Skeleton loading screens
## Description
Add skeleton loading screens for the sites page and the forecast 

## Implementation details 
The skeleton screens can just be a ton of empty divs, laid out in the same format as the elements that will eventually show on the pages. Make sure the styling of the divs (sizing, etc) matches the elements that will eventually show. For animation, tailwind's `animate-pulse` should suffice—read about it in docs [here](https://tailwindcss.com/docs/animation)

## Todos 
- create skeleton screens 
- implement basic css animation
- add to pages, show if `isLoading`

## Links 
- [hubspot blog on skeleton screens](https://blog.hubspot.com/website/skeleton-screens)
- [tailwind animate](https://tailwindcss.com/docs/animation)

neha-vard:
## Description
Add skeleton loading screens for the sites page and the forecast

Fixes #151 

## Screenshots
image
image
image
image


components/dashboard-modules/NumberDisplay.tsx
```diff
@@ -6,6 +6,8 @@ interface Props {
   onClick?: () => void;
 }
 
+const skeleton = `md:my-1.5 text-transparent bg-ocf-gray-1000 w-[100%] rounded-2xl animate-pulse select-none md:leading-none`;
```
AndrewLester: Is there a reason that this skeleton class list has `my:my-1.5` but the other (in SiteCard) doesn't? If there's no need to include it, the class list can be put in a utils file and exported to aid reuse.

EDIT: I see this also is different because of w-[100%] vs w-[50%]. Since these are only applying to <div>s, they should all have display block. Maybe just give the skeleton class flex-1.

components/dashboard-modules/NumberDisplay.tsx
```diff
@@ -15,11 +17,21 @@ const NumberDisplay: FC<Props> = ({ title, value, onClick }) => {
...
     >
-      <div className="mb-2 text-xs md:text-lg text-ocf-gray font-semibold md:font-medium">
-        {title}
-      </div>
-      <div className="text-2xl text-ocf-yellow font-semibold leading-none">
-        {value}
+      <div>
+        <div
+          className={`mb-2 text-xs md:text-lg text-ocf-gray font-semibold md:font-medium ${
+            value == 'Loading...' ? skeleton : ``
```
AndrewLester: Instead of checking `value == 'Loading...'`, maybe this component should explicitly take in a prop called `isLoading` to show this loading state.

AndrewLester (changes requested):
Looks great, one more thing to consider:
- Add `transition-all` as a class to all elements that can have skeleton applied. This way, the skeleton ui fades on/off. Make sure not to put the class _inside_ the skeleton class list, otherwise the transition will also be applied/removed and therefore won't always work correctly.

components/SiteCard.tsx
```diff
@@ -32,37 +34,60 @@ const SiteCard = React.forwardRef<HTMLAnchorElement, SiteCardProps>(
...
-            <h2 className="text-amber text-xl font-semibold">
-              {isLoading ? 'Loading...' : client_site_name ?? 'My Site'}
-            </h2>
-            <div className="flex flex-col mt-2 gap-1">
-              <p className="text-ocf-gray-500 text-xs font-medium">
-                {`Current output: ${
-                  currentOutput != undefined
-                    ? currentOutput.toFixed(2) + ' kW'
-                    : 'loading...'
+            <div className="flex flex-col flex-1 p-4 pl-5">
```
AndrewLester: EDIT: This div (and maybe the one above it) seem to be causing the site card to flow vertically, rather than the horizontal direction it should. Perhaps the `flex-col` is doing it, but just something to check out.

ericcccsliu: Idt its this—isn't helping on my end

ericcccsliu:
great work on the skeleton displays!! for the sites card, though, something in your branch is making them display vertically rather than horizontally—consult the figma for how they're supposed to look.

components/SiteCard.tsx
```diff
@@ -32,37 +34,60 @@ const SiteCard = React.forwardRef<HTMLAnchorElement, SiteCardProps>(
...
+              <h2
+                className={`text-amber text-xl font-semibold ${
+                  isLoading || currentOutput == undefined ? skeleton : ``
```
ericcccsliu: When would `currentOutput` be `undefined`? Wondering if a skeleton would actually be the correct thing to show in this specific case...

ericcccsliu (changes requested):

components/dashboard-modules/SunnyTimeframe.tsx
```diff
@@ -11,12 +11,14 @@ import { hoursToMinutes, millisecondsToHours } from 'date-fns';
...
-  const { forecastData } = useSiteData(siteUUID);
+  const { forecastData, isLoading } = useSiteData(siteUUID);
   const [isRelativeTime, setIsRelativeTime] = useState(false);
   const { timeFormatter } = useDateFormatter(siteUUID);
 
   if (!forecastData) {
-    return <NumberDisplay title="Loading" value="..." />;
+    return (
+      <NumberDisplay title="Loading" value="Loading..." isLoading={isLoading} />
+    );
```
AndrewLester: Though it's hard to tell, you may have noticed that the skeleton effect does not fade out on the `SunnyTimeframe` element, like it does for the other `<NumberDisplay` components on the dashboard. The reason for this is slightly complicated, but it comes down to these lines. Since this component returns early if the `forecastData` doesn't exist, this instance of `<NumberDisplay` will initially be rendered with the skeleton effect. Once the data does exist, this instance will disappear and the one further down will be rendered instead. Despite them being the same component, their rendered HTML elements are different (note that the one below has an `onClick`, which causes the component to show it as a button), and when elements are swapped out on the page they lose their ability to transition CSS wise.

There's a couple ways we could fix this:
- Remove the `onClick` prop for `<NumberDisplay` and don't let them appear as buttons at all (will prevent HTML element swap)
- Remove this return with a `<NumberDisplay` specifically for loading. Instead, move the logic to compute the sunny text and value to another function, and have it consider the case where `forecastData` is undefined, and just return text with "loading". Then, you can always render the bottom `<NumberDisplay` with its `onClick`
- Just add an `onClick` prop to this instance but have it do nothing (`onClick={() => {}}`)

For now, let's go with the last option. Do you think another option is a better long term solution though?

components/dashboard-modules/EnergyRecommendation.tsx
```diff
@@ -38,7 +39,24 @@ const EnergyRecommendation: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
+        justify-center
+        align-center
+        bg-ocf-black-500
+        rounded-2xl
+        h-[100%]"
+      >
+        <div className="bg-ocf-gray-1000 w-[100%] rounded-3xl animate-pulse"></div>
+      </div>
+    );
+  } else if (!isDayTime && currentOutput === 0) {
```
AndrewLester: The `<EnergyRecommendation` component on the dashboard doesn't fade its skeleton layout out similar to the `<SunnyTimeframe` module (read other comment). These lines are the reason in this case, though I'm not sure what the best fixing option is for this one.

For now, let's leave it be and create an issue to track it.

AndrewLester (approved):
Two more things to think about (one is a change), then it's good to merge.

components/dashboard-modules/SunnyTimeframe.tsx
neha-vard: Just committed my changes for the last option. I think that maybe the second option would be a better long term solution.

components/SiteCard.tsx
```diff
@@ -15,8 +15,15 @@ interface SiteCardProps {
...
-      useSiteData(siteUUID);
+    const {
+      forecastData,
+      client_site_name,
+      installed_capacity_kw,
+      isLoading,
+      error,
+    } = useSiteData(siteUUID);
+
+    const noError = error.errors.every((error) => error === undefined);
```
ericcccsliu, line 26: awesome

ericcccsliu (approved):
Awesome work, great job!! Let's just make sure to update those screenshots in your PR description—thank you for including those, by the way—and then you'll be good to merge!!!!

AndrewLester (approved):

PR #190, opened 2023-04-04
Implement Aggregate Dashboard component for displaying the sum of data from all sites on the site list

### LINKED ISSUE #140: Change `/dashboard` route to show aggregate data
## Description

The `/dashboard` route should be modified to show aggregate data for several of a user's sites. If they only have one site, the user should be redirected to `/dashboard/{site's uuid}`. You can use the [useSiteAggregation](https://github.com/openclimatefix/pv-sites-mobile/blob/4e8eea25f43232c50e1b28fed89570c545b53a5b/lib/hooks/useSiteAggregation.ts#L11) hook to get the aggregate data.

## Todos

- [ ] In the `/dashboard/` route page (`/dashboard/index.tsx`), handle the case where the user has 1 site in the `withSites()` call. You can see an example of this inside `/index.tsx`. If they have 0 sites, redirect them to `/dashboard/b97f68cd-50e0-49bb-a850-108d4a9f7b7e` (a test site).
- [ ] Change the dashboard modules that make sense to show on the aggregate dashboard to take in as props a _list_ of site UUIDs, rather than just one. (For example, sunny timeframe doesn't make sense since "how long it'll be sunny for" is an almost per-site quality). Update their internal logic to handle this, and their points of consumption.
- [ ] Create a new `Dashboard` component to be the for an aggregate dashboard.

## Useful links

-  [useSiteAggregation](https://github.com/openclimatefix/pv-sites-mobile/blob/4e8eea25f43232c50e1b28fed89570c545b53a5b/lib/hooks/useSiteAggregation.ts#L11)

rohanvan123:
…tead of a single one

## Description
For this PR, we wanted to configure the `/dashboard` vanilla route to display an aggregation of all the sites. Hence this implied having to alter all the Dashboard modules to now take in an array of siteUUIDs rather than a single one and require all components that we wanted to aggregate to now make use of the useSiteAggregation hook. However, we ran into some issues with the clearsaky data that now should be fixed. 

Fixes #140 

## Todos
- Code Quality


components/graphs/ThresholdGraph.tsx
```diff
@@ -204,15 +202,15 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
    * and returns text indicating increasing/decreasing solar activity
    */
   const getSolarActivityText = () => {
-    if (!graphData) return null;
+    if (!totalExpectedGeneration) return null;
 
-    const currIndex = getCurrentTimeGenerationIndex(graphData);
-    const slope = getTrendAfterIndex(graphData, currIndex);
+    const currIndex = getCurrentTimeGenerationIndex(totalExpectedGeneration);
+    const minMax = getTrendAfterIndex(totalExpectedGeneration, currIndex);
```
AndrewLester: Can we revert this variable name change back to slope?

AndrewLester (approved):

PR #192, opened 2023-04-04
Show tomorrow's threshold graph after dusk

### LINKED ISSUE #187: ThresholdGraph should show tomorrow's graph after dusk
**Describe the bug**
The ThresholdGraph used to show tomorrow's forecast once the time had hit 8pm, and the graph had reached its limit. Now, with it taking its bounds from dawn and dusk calculated by SunCalc, it no longer does this.

**Expected behavior**
After the graph ends (dusk time), tomorrow's forecast should be shown, with tomorrow's range being tomorrow's dawn/dusk times.

**Additional context**
**Ask Kalyn about how to indicate the graph is for tomorrow**. Also, if showing tomorrow's graph, we might want to do something about the way the current time is displayed, since it implies the current time is within the start and end times of the graph, but if we're showing tomorrow, it won't be.

ashayp22:
## Description

- Updates the `useTime()` hook to calculate the next day's dusk and dawn if the current time is after the current day's dusk. This is so the user can view the next day's solar forecast after the current day's dusk.
- Update styling for the Threshold graph based on the above.

Fixes #187

### DIFF
--- a/components/graphs/ThresholdGraph.tsx
+++ b/components/graphs/ThresholdGraph.tsx
@@ -205,21 +205,21 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
     if (!graphData) return null;
 
     const currIndex = getCurrentTimeGenerationIndex(graphData);
-    const minMax = getArrayMaxOrMinAfterIndex(graphData, currIndex);
+    const slope = getArrayMaxOrMinAfterIndex(graphData, currIndex);
 
-    if (minMax) {
-      const { type, index } = minMax;
-      const minMaxForecastDate = timeFormatter.format(
-        new Date(graphData[index].datetime_utc)
+    if (slope) {
+      const { type, endIndex } = slope;
+      const slopeForecastDate = timeFormatter.format(
+        new Date(graphData[endIndex].datetime_utc)
       );
 
       switch (type) {
-        case 'max':
-          return solarIncreasingText(minMaxForecastDate);
-        case 'min':
-          return solarDecreasingText(minMaxForecastDate);
+        case 'increasing':
+          return solarIncreasingText(slopeForecastDate);
+        case 'decreasing':
+          return solarDecreasingText(slopeForecastDate);
         case 'constant':
-          return solarConstantText(minMaxForecastDate);
+          return solarConstantText(slopeForecastDate);
         default:
           return '';
       }
--- a/components/graphs/ThresholdGraph.tsx
+++ b/components/graphs/ThresholdGraph.tsx
@@ -23,7 +23,7 @@ import {
   makeGraphable,
 } from 'lib/graphs';
 
-import { getArrayMaxOrMinAfterIndex } from 'lib/utils';
+import { getTrendAfterIndex } from 'lib/utils';
 
 import { useSiteData } from 'lib/hooks';
 import useDateFormatter from '~/lib/hooks/useDateFormatter';
@@ -205,7 +205,7 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
     if (!graphData) return null;
 
     const currIndex = getCurrentTimeGenerationIndex(graphData);
-    const slope = getArrayMaxOrMinAfterIndex(graphData, currIndex);
+    const slope = getTrendAfterIndex(graphData, currIndex);
 
     if (slope) {
       const { type, endIndex } = slope;
--- a/components/graphs/ThresholdGraph.tsx
+++ b/components/graphs/ThresholdGraph.tsx
@@ -36,7 +36,7 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
   const { currentTime, duskTime, dawnTime } = useTime(latitude, longitude, {
     updateEnabled: timeEnabled,
   });
-  const { timeFormatter } = useDateFormatter(siteUUID);
+  const { timeFormatter, dayFormatter } = useDateFormatter(siteUUID);
 
   const graphData = useMemo(() => {
     if (forecastData && dawnTime && duskTime) {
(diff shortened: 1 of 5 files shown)


lib/utils.ts
```diff
@@ -6,6 +6,7 @@ import {
 import { GetServerSidePropsContext, GetServerSidePropsResult } from 'next';
 import { getCurrentTimeGenerationIndex } from './graphs';
 import { GenerationDataPoint, SiteList } from './types';
+import { start } from 'repl';
```
AndrewLester:
```suggestion

```

lib/utils.ts
```diff
@@ -51,13 +52,19 @@ export const getArrayMaxOrMinAfterIndex = (
...
 
-    const currentDifference = next - current;
+    const currentDifference = curr - prev;
     const currentSlopeSign = Math.sign(currentDifference);
 
     if (firstSlopeSign !== currentSlopeSign && currentSlopeSign !== 0) {
+      // Slope was constant until current index
+      if (firstSlopeSign === 0) {
+        return { type: 'constant', index: startIndex };
+      }
```
AndrewLester: Thanks for adding this

components/graphs/ThresholdGraph.tsx
```diff
@@ -130,14 +130,37 @@ const ThresholdGraph: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
       <div className="flex flex-row justify-between">
-        <p className="text-white text-xs font-medium ml-6">{startTime}</p>
-        <p className="text-white text-xs font-medium mr-6">{endTime}</p>
+        <div>
+          <p className="text-white text-xs sm:text-base font-semibold sm:font-medium ml-3 sm:ml-6">
+            {startTime}
+          </p>
+          <p className="text-white text-xs sm:text-base font-normal ml-3 sm:ml-6">
+            {startDay}
+          </p>
```
AndrewLester, line 150: Can we only show the day if you're looking at tomorrow's forecast?

AndrewLester (approved):
Great work, thanks for fixing some other stuff as well.

components/graphs/ThresholdGraph.tsx
ashayp22: Any reason why not to show the time?

lib/hooks/useTime.ts
```diff
@@ -24,21 +24,32 @@ const useTime = (
...
   const [currentTime, setCurrentTime] = useState(Date.now());
 
+  /**
+   * @returns 12:00 AM GMT-05 of the next day
+   */
+  const getNextDay = () => {
+    const now = new Date();
+    now.setDate(now.getDate() + 1);
+    now.setHours(0, 0, 0, 0);
+    return now;
```
ericcccsliu: tiny detail but rename this away from `now` since it'll represent the next day

lib/utils.ts
```diff
@@ -51,13 +51,19 @@ export const getArrayMaxOrMinAfterIndex = (
   startIndex += 1;
```
ericcccsliu, line 51: actually refers to lines 38-43, but we should probably change `type` returned in the `startIndex === array.length - 1` case to `constant`

lib/utils.ts
```diff
@@ -20,7 +20,7 @@ export function camelCaseID(id: string) {
 }
 
 interface MinMaxInterface {
-  type: 'min' | 'max';
+  type: 'min' | 'max' | 'constant';
   index: number;
 }
```
ericcccsliu: This change is great, but it also changes the responsibility of this function quite a bit. Seems like now, we're basically returning the graph's trend by adding the possibility of the graph being `constant`, and deviating from the behavior of just returning the min or max. Accordingly, I think we should refactor this function (and this interface) to just return the trend of the graph, whether it's increasing, decreasing, or constant. 

```TypeScript
interface GraphTrendInterface {
   behavior: 'increasing' | 'decreasing' | 'constant';
   endIndex: number;
}
```

components/graphs/ThresholdGraph.tsx
AndrewLester: Ah my language was confusing, sorry. I mean that we should not show the day if we're looking at today's forecast.

ashayp22: Why not include the day? It can strengthen the yellow marker on the graph's indication that we are showing the current day's forecast.

ericcccsliu (changes requested):
Should probably resolve/address the comments I left before merging

components/graphs/ThresholdGraph.tsx
AndrewLester: I'm for as little text as possible on the dashboard. That way, the text that is there is all easy to parse and notice without having to filter redundant info. I think it is assumed that the graph's start and end are for the current day if the yellow marker is there. We can ask Kalyn about this though.

lib/utils.ts
```diff
@@ -34,11 +34,11 @@ interface MinMaxInterface {
 export const getArrayMaxOrMinAfterIndex = (
```
ericcccsliu:
```suggestion
export const getTrendAfterIndex = (
```
or something like that, just need to change this name accordingly

lib/utils.ts
```diff
@@ -19,9 +19,9 @@ export function camelCaseID(id: string) {
   return [first, ...rest.map(capitalize)].join('');
 }
 
-interface MinMaxInterface {
-  type: 'min' | 'max';
-  index: number;
+interface SlopeInterface {
+  type: 'increasing' | 'decreasing' | 'constant';
+  endIndex: number;
 }
```
ericcccsliu, line 25: thanks for the changes!

ericcccsliu (approved):
great work—literally just one tiny comment left, and then you're good to merge anytime! thanks for putting this up!!

PR #194, opened 2023-04-05
Add animations

### LINKED ISSUE #198: Linear gradient not working intermittently on mobile
**Describe the bug**
The linear gradient element that we have under our threshold and sites graphs is occasionally not working, displaying all black, on mobile browsers. 
**To Reproduce**

1. Go to the sites page or dashboard page on Safari or Chrome on your mobile device
2. Reload a couple of times 
3. See error

**Expected behavior**
Should look like the desktop views. 

**Additional context**
This is likely due to us using the `transparent` keyword, which may be unsupported on mobile. Changing transparent values to `rgba(255,255,255,0)` should fix this. Found this [apple css guide](https://developer.apple.com/library/archive/documentation/InternetWeb/Conceptual/SafariVisualEffectsProgGuide/Gradients/Gradient.html) that uses `rgba` rather than transparent.

AndrewLester:
## Status:

🚀 Ready

## Description

This does a couple things I'm not too sure about:
- Upgrades to next 13 and React 18 to support framer-motion (from what I could tell this was the easiest way)
- Fixes issues with `<Image` due to this upgrade
- Adds page transitions following: https://letsbuildui.dev/articles/animated-page-transitions-in-nextjs (mobile only)

The transition in here doesn't look too great in my opinion, especially since we're navigating with a bottom nav. Maybe we want to go with a simple fade in the long run. Or nothing at all, since the skeleton screens will look good enough.

Fixes #<number>


next.config.js
```diff
@@ -1,6 +1,9 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
+  experimental: {
+    newNextLinkBehavior: false,
```
AndrewLester, line 5: I decided not to deal with fixing our usages of `<Link` (to be inline with Next 13) in this PR, but I probably should.

ericcccsliu:
discussed changing transitions to the x-axis, transitions only on increasing depth, and a bottom navbar with just "back".

ericcccsliu (approved):

PR #199, opened 2023-04-06
Add weather component to dashboard

### LINKED ISSUE #168: Add weather icons to dashboard to show weather at a glance
## Description
Add a component to the dashboard to display the weather at a glance 
Screen Shot 2023-03-28 at 1 45 06 AM

## Implementation Details
We'll be determining which icon to show based on the total expected output over the entire day. This can be calculated using the `getTotalExpectedOutput` function: 
https://github.com/openclimatefix/pv-sites-mobile/blob/27fa0d76aa78b4a649b0dd8873037dd9c1e78637/components/dashboard-modules/ExpectedTotalOutput.tsx#L6-L24
which should probably be refactored into a `utils` file. Furthermore, we'll need to calculate this over each day—we can do this using the `ForecastDataOverDateRange` function: 
https://github.com/openclimatefix/pv-sites-mobile/blob/27fa0d76aa78b4a649b0dd8873037dd9c1e78637/lib/graphs.ts#L39-L49
To find out how "sunny" it'll be, we need to also calculate the total expected output of each day under a clear sky, so we'll have to make sure those functions work with the clearsky estimate, and duplicate+adjust them if they don't. Dividing the forecast expected output and the clearsky expected output should give us a fraction, which we can compare against arbitrary constants to determine which icons are shown. 

## Todos
- [x] calculate output of clear sky day
- [x] scale each day against their respective clear sky days 
- [x] create arbitrary breakpoints to determine which icon must be shown 
- [ ] create/style component according to [figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=1096%3A3089&t=i6hBykmPGHh2S3r1-1)

## Links 
- [figma](https://www.figma.com/file/Y2IIxfo496bwxFsz8zo9hI/Wireframes?node-id=1096%3A3089&t=i6hBykmPGHh2S3r1-1)

anyaparekh:
## Status:
:rocket: Ready

## Description
* Use useTime hook and GetExpectedOutputOverDay to render three-day visual weather forecast
* Replace percent yield component with weather component
Closes #168 

## Screenshots
image


ericcccsliu:
Screen Shot 2023-04-06 at 7 15 32 PM
hmm why does this say "today" twice

components/WeatherCard.tsx
```diff
@@ -0,0 +1,109 @@
+import { FC } from 'react';
+import { useSiteData } from '~/lib/hooks';
+import useTime from '~/lib/hooks/useTime';
+
+import CloudyIcon from './icons/CloudyIcon';
+import SunnyIcon from './icons/SunnyIcon';
+import PartlyCloudyIcon from './icons/PartlyCloudyIcon';
```
ericcccsliu: Instead of importing these separately, we should add these as exports into /icons/index.tsx! Then, we'll be able to import these all in one line.

components/WeatherCard.tsx
```diff
@@ -0,0 +1,109 @@
...
+  );
+
+  return [
+    getTotalExpectedOutput(forecastData),
+    getTotalExpectedOutput(forecastData) / getTotalExpectedOutput(clearSkyData),
+    newDuskTime.getDay(),
+  ];
+};
+
+const getWeatherDisplay = (diff: number, total: number, day: number) => {
```
ericcccsliu: can we change `diff` to `clearskyCapacity`?

components/WeatherCard.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const dates = ['Today', 'Mon', 'Tues', 'Wed', 'Thurs', 'Fri', 'Sat', 'Sun'];
```
ericcccsliu: This + its usage doesn't match up with the indices returned by `getDay()`

components/WeatherCard.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const GetExpectedOutputOverDay = (
+  duskTime: Date,
+  dawnTime: Date,
+  offset: number,
+  forecastValues: GenerationDataPoint[],
+  clearSkyEstimate: GenerationDataPoint[]
+) => {
+  const newDuskTime = new Date(duskTime.getTime() + offset * 3600000);
+  const newDawnTime = new Date(dawnTime.getTime() + offset * 3600000);
```
ericcccsliu: can we make a constant above all this called `millisInHour`, set it equal to 3600000?

AndrewLester: This can also be imported from `date-fns` as a constant I think.

ericcccsliu (approved):
LGTM

PR #219, opened 2023-04-17
Use Day.js and move files

### LINKED ISSUE #209: Replace `date-fns` with `day.js`
## Description

`date-fns` doesn't have the features we need to work with timezones, and we end up using mostly builtin date methods anyway. `Day.js`'s [setDefaultTimezone](https://day.js.org/docs/en/timezone/set-default-timezone) will be very useful in comparing dates/days of week in the site timezone:
https://github.com/openclimatefix/pv-sites-mobile/blob/f6a4accbac478e71a2f304775d322dad0a9a5058/components/graphs/ThresholdGraph.tsx#L244-L249

## Todos

- [ ] Todo

## Useful links

- Useful link

AndrewLester:
## Status:

:rocket: Ready

## Description

**Waiting on #226 and #231**

This PR moves a lot of files and also uses Day.js instead of `date-fns`. Here's a quick overview:
- `components` folder is moved into `lib`, so that components and relevant library functions can be looked at nearer to each other. This means we don't have to scroll over the data/content folders when working on both of these relevant folders.
- `date-fns` is uninstalled and replace with `Day.js`. `Day.js` makes it easier to translate times into a different timezone, and also has the same functionality as `date-fns`
- 

Fixes #<number>


lib/components/Dashboard.tsx
```diff
@@ -1,25 +1,23 @@
...
-  const sites = siteUUIDs.site_list.map((site) => site.site_uuid);
-  const { client_site_name } = useSiteData(sites[0]);
+const Dashboard: FC<DashboardProps> = ({ sites }) => {
+  const isAggregate = sites.length > 1;
+
   return (
     <div className="bg-ocf-black max-w-screen-xl w-screen min-h-screen px-4 mb-[var(--bottom-nav-margin)]">
       <h1 className="mt-4 mb-4 text-ocf-gray text-3xl font-bold">
-        {isAggregate ? 'Aggregate Dashboard' : client_site_name}
+        {isAggregate ? 'Dashboard' : sites[0].client_site_name}
```
ericcccsliu: need to think of a better distinction in name

lib/components/Dashboard.tsx
```diff
@@ -29,33 +27,33 @@ const Dashboard: FC<DashboardProps> = ({ siteUUIDs }) => {
         </div>
         {!isAggregate && (
           <div className="grid-in-Sunny">
-            <SunnyTimeframe siteUUID={sites[0]} />
+            <SunnyTimeframe site={sites[0]} />
```
ericcccsliu: need to also think about a better way to do this lol

lib/components/Dashboard.tsx
```diff
@@ -29,33 +27,33 @@ const Dashboard: FC<DashboardProps> = ({ siteUUIDs }) => {
...
             Metrics
           </h2>
         </div>
         <div className="grid-in-Expected-Total-Output">
-          <ExpectedTotalOutput siteUUIDs={sites} />
+          <ExpectedTotalOutput sites={sites} />
         </div>
         <div className="grid-in-Current-Output md:hidden">
-          <CurrentOutput siteUUID={sites[0]} />
+          <CurrentOutput site={sites[0]} />
```
ericcccsliu: currentOutput can take in multiple sites

ericcccsliu (approved):

lib/components/dashboard-modules/CurrentOutput.tsx
```diff
@@ -1,13 +1,17 @@
...
-    ? getCurrentTimeGeneration(forecastData.forecast_values)
-    : undefined;
+interface CurrentOutputProps {
+  site: Site;
+}
+
+const CurrentOutput: FC<CurrentOutputProps> = ({ site }) => {
+  const { forecastData, isLoading } = useSiteData(site.site_uuid);
+  const currentOutput =
+    forecastData && getCurrentTimeGeneration(forecastData.forecast_values);
```
ericcccsliu: could we aggregate together current output like in the other modules?

lib/components/dashboard-modules/SunnyTimeframe.tsx
```diff
@@ -1,24 +1,18 @@
...
+const SunnyTimeframe: FC<{ site: Site }> = ({ site }) => {
+  const { forecastData, isLoading } = useSiteData(site.site_uuid);
   const [isRelativeTime, setIsRelativeTime] = useState(false);
-  const { timeFormatter } = useDateFormatter(siteUUID);
+  const { timezone } = useSiteTime(site);
 
-  const thresholdCapacityKW = installed_capacity_kw
-    ? installed_capacity_kw * graphThreshold
-    : 0;
+  const thresholdCapacityKW = site.installed_capacity_kw * graphThreshold;
```
ericcccsliu: i think it's fine to make this use multiple sites too

lib/components/dashboard-modules/SunnyTimeframe.tsx
```diff
@@ -46,23 +40,23 @@ const SunnyTimeframe: FC<{ siteUUID: string }> = ({ siteUUID }) => {
...
     let timeUnit = difference === 1 ? 'Hour' : 'Hours';
 
     // If we are under an hour, round to minutes
     if (difference < 1) {
-      difference = hoursToMinutes(difference);
+      difference = dayjs.duration(difference, 'hours').asMinutes();
       timeUnit = difference ? 'Minute' : 'Minutes';
     }
 
     value = `${difference} ${timeUnit}`;
```
ericcccsliu: i think it's here? or might be somewhere else. need a toFixed or a round.

Screen Shot 2023-04-17 at 4 21 02 PM

lib/components/graphs/ThresholdGraph.tsx
```diff
@@ -18,51 +18,48 @@ import {
...
+} from '~/lib/generation';
+import { timeFormat, weekdayFormat } from '~/lib/format';
+import dayjs from 'dayjs';
+
+interface ThresholdGraphProps {
+  sites: Site[];
+}
+
+const ThresholdGraph: FC<ThresholdGraphProps> = ({ sites }) => {
+  const representativeSite = sites[0];
```
ericcccsliu, line 41: For the case with multiple sites, I don't know if showing a representative site would be correct. Instead, it might be better to just show this in the user's time zone.

ericcccsliu: can discuss more.

lib/components/graphs/Graph.tsx
```diff
@@ -69,24 +63,22 @@ function getXTickValues(times: number[], numTicks: number) {
...
 
-const Graph: FC<{ siteUUIDs: string[] }> = ({ siteUUIDs }) => {
-  // TODO: we want a aggregate form of the clearskyData variable
-  const { latitude, longitude } = useSiteData(siteUUIDs[0]);
+interface GraphProps {
+  sites: Site[];
+}
+
+const Graph: FC<GraphProps> = ({ sites }) => {
+  const representativeSite = sites[0];
```
ericcccsliu, line 100: same issue with using a representative site here, and in other places probably

lib/components/dashboard-modules/SunnyTimeframe.tsx
AndrewLester: I'll hold off on this PR til @rohanvan123 is done then, since he's making the changes to support multiple sites in this components.

data/site-list.json
```diff
@@ -60,7 +60,7 @@
       "tilt": null,
       "latitude": 50.0,
       "longitude": 50.0,
-      "installed_capacity_kw": 4,
+      "installed_capacity_kw": 0,
```
ericcccsliu: wait why did these get changed to 0?

ericcccsliu (commented):
wait w

data/site-list.json
AndrewLester: This is because this sites never load a forecast since there's no data for them, and to avoid confusion in the way the aggregate site capacity is calculated, being that it sums the several sites, I set these to 0. That way, the threshold, like the forecast values, both stay the same from the aggregate to single site dashboard (for the one that works)

PR #226, opened 2023-04-20
API Auth0

### LINKED ISSUE #224: Correctly setup SWR to send Auth0 Token
## Description

Given that the `pv-site-api` is now protected by auth, we need to attach the user's Auth0 token with all requests we make to the API to be authorized. To do this, we'll want to replace the default and one-off SWR "fetchers" we use throughout the app to attach the token. Some example code using `axios` can be found in the `nowcasting` repo, though we will use `fetch`:

https://github.com/openclimatefix/nowcasting/blob/d72e272acc93ab7cbdb6af01d54d5376f7de2b25/apps/nowcasting-app/components/helpers/utils.ts#L110-L129

We have a similar endpoint to get the user's Auth0 token.

## Todos

- [ ] Change the default and one-off fetchers to attach the user's Auth0 token
- [ ] Store the Auth0 token in memory so it doesn't have to be `fetch`'d before every request.

## Useful links

- https://swr.vercel.app/blog/swr-v1.en-US#no-more-default-fetcher

AndrewLester:
## Status:

:rocket: Ready

## Description

Tested this locally with the local API and database. Everything seems to work.

Fixes #224

### DIFF
--- a/env.d.ts
+++ b/env.d.ts
@@ -7,7 +7,7 @@ namespace NodeJS {
     AUTH0_ISSUER_BASE_URL: string;
     AUTH0_CLIENT_ID: string;
     AUTH0_CLIENT_SECRET: string;
-    AUTH0_AUDIENCE;
+    AUTH0_AUDIENCE: string;
     NEXT_PUBLIC_MAPBOX_PUBLIC: string;
     NEXT_PUBLIC_AUTH0_LOGOUT_REDIRECT: string;
     NEXT_PUBLIC_API_BASE_URL_POST: string;
--- a/env.d.ts
+++ b/env.d.ts
@@ -7,6 +7,7 @@ namespace NodeJS {
     AUTH0_ISSUER_BASE_URL: string;
     AUTH0_CLIENT_ID: string;
     AUTH0_CLIENT_SECRET: string;
+    AUTH0_AUDIENCE;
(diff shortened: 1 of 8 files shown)


pages/api/auth/[...auth0].ts
```diff
@@ -1,3 +1,17 @@
...
 
-export default handleAuth();
+export default handleAuth({
+  async login(req, res) {
+    try {
+      await handleLogin(req, res, {
+        authorizationParams: {
+          audience: process.env.AUTH0_AUDIENCE,
+          scope: 'openid profile email offline_access',
+          useRefreshTokens: true,
```
Comment (AndrewLester) [inline pages/api/auth/[...auth0].ts:10]: This stuff is required to get an access token that can be verified by the API

env.d.ts
```diff
@@ -7,6 +7,7 @@ namespace NodeJS {
     AUTH0_ISSUER_BASE_URL: string;
     AUTH0_CLIENT_ID: string;
     AUTH0_CLIENT_SECRET: string;
+    AUTH0_AUDIENCE;
```
ericcccsliu: does this default to `any`? should it?

ericcccsliu (approved):

env.d.ts
AndrewLester: Yep, and no. I'll fix this

PR #242, opened 2023-04-24
Add inverter data to useSiteData hook

### LINKED ISSUE #238: Add inverter data to `useSiteData` hook
## Description

We need to get the inverter data for all the inverters for a site with the `useSiteData` hook. This will use the new endpoint we just created to get the inverter data for a site:

https://github.com/openclimatefix/pv-site-api/blob/9561afc643a1d06fb1a6fac23964d0ae593cbb50/pv_site_api/main.py#L502-L524

## Todos

- [x] Add a `useSWR` query to get this data from the API
- [x] Perhaps add some TypeScript types to represent this data in the frontend. There may be some almost done examples in the `enode.ts` file.
- [ ] sum up current actual generations from all inverters associated with the site 
- [x] Return the data from the `useSiteData` hook

neha-vard:
## Description
Get the inverter data for all the inverters for a site with the useSiteData hook.

Fixes #238


lib/enode.ts
```diff
@@ -16,19 +16,35 @@ export async function clearUsers(userIDs: string[]) {
...
+  };
+  information: {
+    id: string;
+    brand: string;
+    model: string;
+    siteName: string;
+    installationDate: string;
+  };
+  location: {
+    longitute: number | null;
```
ericcccsliu:
```suggestion
    longitude: number | null;
```

lib/enode.ts
```diff
@@ -16,19 +16,35 @@ export async function clearUsers(userIDs: string[]) {
...
+    latitude: number | null;
   };
 };
+
+export type Inverters = {
+  inverters: Inverter[];
+};
```
ericcccsliu: Reading through the docs, it seems like this type (at least if it's being used as the return type for the inverters endpoint) should just be equal to `Inverter[]`—there shouldn't be a key called `inverters`

neha-vard: That makes sense! In the endpoint, I think we return Inverters(inverters=inverters), where inverters is a key. Would this need to change in the endpoint, or would it still work if I just used Inverter[] as the type in the invertersFetcher?

lib/api.ts
```diff
@@ -93,6 +94,11 @@ export const actualsFetcher: Fetcher<ActualData> = async (url: string) => {
   return parseActualData(unparsedData);
 };
 
+export const invertersFetcher: Fetcher<Inverters> = async (url: string) => {
+  const inverterData: Inverters = await fetcher(url);
+  return inverterData;
+};
+
```
AndrewLester: Since this fetcher mostly just turns the URL into JSON, we can actually just use the default fetcher we've supplied to SWR. More on this in another comment below.

The other thing this fetcher does is it appropriately sets the `type` for the data, but this can also be done in another way.
```suggestion
```

lib/sites.ts
```diff
@@ -52,6 +53,11 @@ export function useSiteData(siteUUID: string) {
     actualsFetcher
   );
 
+  const { data: inverterData, error: inverterError } = useSWR(
+    `${process.env.NEXT_PUBLIC_API_BASE_URL_GET}/sites/${siteUUID}/inverters`,
+    invertersFetcher
```
AndrewLester: If you don't supply the fetcher here, SWR will use the default one (that we supplied in `_app.tsx`). It performs the same logic, so using the default is ok.
```suggestion

```

lib/sites.ts
```diff
@@ -52,6 +53,11 @@ export function useSiteData(siteUUID: string) {
     actualsFetcher
   );
 
+  const { data: inverterData, error: inverterError } = useSWR(
```
AndrewLester: Since you're not using a custom fetcher here anymore, you can tell swr to set the type of `data` inline:
```suggestion
  const { data: inverterData, error: inverterError } = useSWR<Inverters>(
```

AndrewLester (approved):

lib/enode.ts
ericcccsliu: ok, sounds good, you're correct then!

ericcccsliu (approved):
LGTM, great work

PR #249, opened 2023-04-25
Inverter Page

### LINKED ISSUE #218: Build inverter view/select page
## Description

Build out the page that users will be redirected to after initially linking their inverters with Enode. 
- Because Enode seems to only support linking all inverters of _just one brand_ at a time, we need some functionality to add more inverters from different brands 
- After a user doesn't want to add more inverters, we need to bring them to a page where they may select a subset of inverters which correspond to their site 
- These inverters need to be sent to the API via a `POST`—this hasn't really been planned in the API yet, so hold off on this for now, but make sure that we can easily add the inverters to a `POST` request. 

Screen Shot 2023-04-16 at 11 30 10 PM
Screen Shot 2023-04-16 at 11 37 37 PM

## Todos

- [x] Create a selectable card component for individual inverters 
- [x] Create inverter view/select pages 
- [x] Create "add more inverters" button that redirects users back to Enode's Link UI, using the `/enode/link` endpoint 
- [x] Populate inverter cards using API, specifically using the `/inverters` endpoint 
- [x] Store selected inverter UUIDs in context to prepare for `POST`ing

## Time estimate
tbd

ericcccsliu:
## Status:
:rocket: Ready

## Description
Gets all the UI up. Preliminary PUT support. Adds fake inverters to mock API. You can access the inverter pages here: 
[http://localhost:3000/inverters/725a8670-d012-474d-b901-1179f43e7182](http://localhost:3000/inverters/725a8670-d012-474d-b901-1179f43e7182) when running locally. 

Fixes #218

## Screenshots
Screen Shot 2023-04-25 at 11 09 27 PM
Screen Shot 2023-04-25 at 11 09 58 PM
Screen Shot 2023-04-25 at 11 10 29 PM
Screen Shot 2023-04-25 at 11 11 21 PM


data/inverters.json
AndrewLester, line 1: Just realized this is faking our new API endpoints, pretty cool

pages/inverters/[uuid].tsx
```diff
@@ -0,0 +1,56 @@
...
+        lastPageCallback={() => setPage(Page.View)}
+      />
+    </div>
+  );
+};
+
+export const getServerSideProps = withSites({
+  async getServerSideProps(ctx) {
+    const { sites, query } = ctx;
+    if (!sites.map((site) => site.site_uuid).includes(query.uuid as string)) {
```
Comment (AndrewLester) [inline pages/inverters/[uuid].tsx:48]: Perhaps a call to `find` would be more intuitive? (not sure if correct syntax)
```suggestion
    if (!sites.find((site) => site.site_uuid === query.uuid as string)) {
```

pages/inverters/[uuid].tsx
```diff
@@ -0,0 +1,56 @@
...
+      return {
+        notFound: true,
```
Comment (AndrewLester) [inline pages/inverters/[uuid].tsx:50]: Should we think about 404 page designs?

lib/components/form/ViewInverters.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  const nextPageOrSubmit = async () => {
+    if (isSelectMode) {
+      if (!didSubmit) setDidSubmit(true);
+      await trigger({ client_ids: selectedInverters });
+    }
+    nextPageCallback();
+  };
+
+  const addOrRemove = (inverter: string) => {
```
AndrewLester: Rename, toggle?
```suggestion
  const toggle = (inverter: string) => {
```

lib/components/form/ViewInverters.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <div />
+        )}
+        <Button
+          variant="solid"
+          disabled={(isSelectMode && selectedInverters.length < 1) || didSubmit}
+          onClick={nextPageOrSubmit}
+        >
+          {didSubmit && <Spinner width={5} height={5} margin={4} />}
+          {isSelectMode ? 'Finish' : 'Next'}
+          {didSubmit && <div className="mx-4 w-5" />}
```
AndrewLester, line 145: I feel like these empty divs may be more clutter than trying to correct this with CSS... unsure if it's easy though.

lib/components/form/ViewInverters.tsx
```diff
@@ -0,0 +1,152 @@
...
+  return fetch(url, {
+    method: 'PUT',
+    body: JSON.stringify(arg),
+    credentials: 'include',
+    headers: {
+      'Content-Type': 'application/json',
+      ...options.headers,
+    },
+  });
+}
```
AndrewLester, line 24: Do we want to move this function into some other file? Seems re-usable

lib/components/form/SiteDetails.tsx
```diff
@@ -33,7 +30,7 @@ const SiteDetails: FC<SiteDetailsProps> = ({ site }) => {
 
   const nextPageCallback = () => {
     if (page === Page.Details) {
-      router.push(mobile ? '/sites' : '/dashboard');
+      router.push('/dashboard');
```
AndrewLester: Is there a reason for removing this? I added it since desktop shouldn't use the sites page.

lib/components/form/SiteDetails.tsx
```diff
@@ -65,14 +62,10 @@ const SiteDetails: FC<SiteDetailsProps> = ({ site }) => {
...
   return (
     <div className="w-full md:flex-col md:justify-center">
-      <div className="flex h-[var(--nav-height)] w-full flex-row items-center justify-between bg-ocf-black px-5 md:justify-center md:py-2">
-        <div className="md:hidden">
-          {!(page === Page.Location && sites.length === 0) && (
-            <BackButton onClick={lastPageCallback} />
-          )}
-        </div>
-        <NowcastingLogo />
-      </div>
```
AndrewLester, line 75: Is this stuff also supposed to be removed?

lib/components/InverterCard.tsx
```diff
@@ -1,28 +1,57 @@
...
     </div>
-  </div>
-);
+  );
+
+  return selectMode ? (
+    <button
+      onClick={() => {
+        !!onClick && onClick();
+      }}
```
AndrewLester:
```suggestion
      onClick={onClick}
```

lib/components/form/ViewInverters.tsx
```diff
@@ -0,0 +1,152 @@
...
+          )}
+          <Button
+            variant="solid"
+            disabled={isSelectMode && selectedInverters.length < 1}
+            hidden="md"
+            className="mb-8 mt-auto w-full self-center"
+            onClick={nextPageOrSubmit}
+          >
+            {isSelectMode ? 'Finish' : 'Next'}
+          </Button>
```
AndrewLester, line 127: Duplicating buttons on these pages seems to be the only way to do it with the differing styles and positioning... but the duplication on these files just makes them long and a bit harder to read. Maybe we can think about this later. Seems like the `hidden` attribute makes it cleaner at least.

lib/components/Button.tsx
```diff
@@ -11,22 +13,26 @@ const Button: FC<PropsWithChildren<Props>> = ({
...
   return (
     <button
       suppressHydrationWarning
       form={form}
       onClick={onClick}
       disabled={disabled}
-      className={variant === 'solid' ? solidButton : outlinedButton}
+      className={`${variant === 'solid' ? solidButton : outlinedButton} ${
+        hidden && `${hidden}:hidden`
+      } ${className && className}`}
```
AndrewLester: You could default this variable to `''` in the parameters if it doesn't type check with just below
```suggestion
      } ${className}`}
```

AndrewLester (approved):

lib/components/form/SiteDetails.tsx
ericcccsliu: Yep, replaced by `<BackNav>`

lib/components/form/ViewInverters.tsx
ericcccsliu: Yeah, it does seem reusable, but I don't really want to spend time to deal with passing in different types for `arg` right now...

lib/components/form/ViewInverters.tsx
ericcccsliu: idk, just makes it really easy to center the middle element right now

pages/inverters/[uuid].tsx
Comment (ericcccsliu) [reply pages/inverters/[uuid].tsx:50]: yes definitely

lib/components/form/SiteDetails.tsx
ericcccsliu: yeah no i just got confused gonna put it back

PR #265, opened 2023-04-28
Go back to previously visited dashboard

### LINKED ISSUE #256: Have dashboard page remember the last dashboard you were on
## Description
Right now, the dashboard page does not go back to the previous dashboard you selected. If you click on a site through the sites sidebar (problem manifests on desktop especially) and click back to more info or something, and then return to dashboard, the aggregate dashboard shows. 

This might be a place for context, or even an issue for local storage (dashboard selection would persist under reload). @AndrewLester @ashayp22 what do y'all think?

ashayp22:
## Description

Users can navigate to the previously visited dashboard using the Dashboard link in the <Navbar/> through `useContext()`.

Fixes #256


lib/provider.tsx
ericcccsliu, line 1: `provider.tsx` seems very general for what this is—could we rename this to something more specific to the dashboard?

lib/provider.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useState, useContext, FC, PropsWithChildren } from 'react';
+import { App } from './types';
+
+const AppContext = React.createContext<App | null>(null);
```
ericcccsliu: Could we also rename this to something more specific to the dashboard? Again, `AppContext` is too general

lib/types.ts
```diff
@@ -142,3 +142,8 @@ export type Inverters = {
 export type InverterPutData = {
   client_ids: string[];
 };
+
+export type App = {
```
ericcccsliu: we should also rename this—too many things called `App` rn

ericcccsliu (approved):
great work tysm

PR #277, opened 2023-04-29
Implement correct tomorrow behavior on threshold graph

### LINKED ISSUE #259: Implement correct tomorrow behavior on threshold graph
## Description

Currently, our threshold graph is only displaying tomorrow's forecast after midnight. We want it to start displaying tomorrow's forecast after sunset today.

neha-vard:
## Description
Currently, our threshold graph is only displaying tomorrow's forecast after midnight. We want it to start displaying tomorrow's forecast after sunset today.

Fixes #259

### DIFF
--- a/lib/time.ts
+++ b/lib/time.ts
@@ -28,9 +28,12 @@ export const useSiteTime = (
     latitude && longitude ? ['/timezone', longitude, latitude] : null,
     () => fetchTimeZone(latitude, longitude)
   );
+
   const [currentTime, setCurrentTime] = useState(dayjs().tz(timezone));
+  const [tomorrow, setTomorrow] = useState(currentTime.add(1, 'day'));
 
   useEffect(() => setCurrentTime(dayjs().tz(timezone)), [timezone]);
+  useEffect(() => setTomorrow(currentTime.add(1, 'day')), [currentTime]);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


ericcccsliu (approved):
wooooooooooooooooo

amazing work

lib/time.ts
```diff
@@ -28,9 +28,12 @@ export const useSiteTime = (
     latitude && longitude ? ['/timezone', longitude, latitude] : null,
     () => fetchTimeZone(latitude, longitude)
   );
+
   const [currentTime, setCurrentTime] = useState(dayjs().tz(timezone));
+  const [tomorrow, setTomorrow] = useState(currentTime.add(1, 'day'));
 
   useEffect(() => setCurrentTime(dayjs().tz(timezone)), [timezone]);
+  useEffect(() => setTomorrow(currentTime.add(1, 'day')), [currentTime]);
```
AndrewLester: Since `tomorrow` is always just a value computed from `currentTime`, it doesn't have to be a state variable.

You _could_ `useMemo` this, but maybe not necessary.
```suggestion
  const tomorrow = currentTime.add(1, 'day');

  useEffect(() => setCurrentTime(dayjs().tz(timezone)), [timezone]);
```

AndrewLester (approved):
Looks great

PR #284, opened 2023-04-30
Design an algorithm to determine the moving average of our forecast array to be used for energy predictions

### LINKED ISSUE #267: implement moving average for graph increasing or decreasing
## Description
Right now, our "increasing until.../decreasing until..." text is wholly sensitive to small local variations. Implement a moving average function on the data, which averages each data point with the _n_ data points to its right and left. Do it in linear time, constant 1 because that's fun haha.

## Links
- [moving average Wikipedia (fun read)](https://en.wikipedia.org/wiki/Moving_average)

rohanvan123:
## Description
For this issue I wrote a method to take a list of generation data points and comp[ute the moving average. Essentially for each index it take the minimum of the number of elements before, the number of elements after, and the $\lfloor \frac{period} 2 \rfloor$ and computes the average for that specific range. We then use this new array to determine the solar activity text for the threshold graph. 

Fixes #267

### DIFF
--- a/lib/components/graphs/ThresholdGraph.tsx
+++ b/lib/components/graphs/ThresholdGraph.tsx
@@ -19,6 +19,7 @@ import {
 } from '../icons/FutureThreshold';
 
 import {
+  SimpleMovingAverage,
   addTimePoint,
   getTrendAfterIndex,
   graphThreshold,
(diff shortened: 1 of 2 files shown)


lib/components/graphs/ThresholdGraph.tsx
```diff
@@ -51,6 +52,10 @@ const ThresholdGraph: FC<ThresholdGraphProps> = ({ sites }) => {
       updateEnabled: timeEnabled,
     });
 
+  if (totalForecastedGeneration) {
+    console.log(SimpleMovingAverage(totalForecastedGeneration, 5));
+  }
+
```
AndrewLester:
```suggestion
```

AndrewLester (approved):

PR #286, opened 2023-04-30
Update save change button to make PUT request only 

### LINKED ISSUE #276: Implement PUT after editing sites
## Description

PUT to api PUT endpoints in h4i/enode branch. Can be done after every hit of "save changes." Needs to be done for both inverters and site details

Related: https://github.com/openclimatefix/pv-sites-mobile/blob/3255308e1a1e86a71cb0a1df192833e0fc37a4a2/lib/components/form/ViewInverters.tsx#L70

jackypark9852:
- Updates the save changes button to send the  PUT request and display check mark when complete 
- Make minor UI changes for site details and inverters page
## Status:
:rocket: Ready

## Description

Fixes #276 

## Screenshots
image
image


AndrewLester (approved):

lib/components/form/ViewInverters.tsx
```diff
@@ -47,17 +49,26 @@ const ViewInverters: FC<ViewInvertersProps> = ({
...
+      await submit();
     }
     nextPageCallback();
   };
 
+  const submit = async () => {
+    if (!didSubmit) setDidSubmit(true);
+    await trigger(selectedInverters);
+    setShowSuccessIcon(true);
+    setDidSubmit(false);
```
ericcccsliu, line 67: is there a reason why we're setting `didSubmit` back to `false` here?

ericcccsliu: ah ok i get it it's to allow for multiple edits

ericcccsliu: Could we move this handling into the `options` parameter of `useSWRMutation`? That way we can also more specifically define behavior on success and on error—right now, it seems like a success icon might be shown regardless of if the `PUT` goes through.

ericcccsliu (approved):

PR #287, opened 2023-04-30
Hide nav on link page

AndrewLester:
## Description

Fixes #<number>

### DIFF
--- a/lib/components/Layout.tsx
+++ b/lib/components/Layout.tsx
@@ -12,7 +12,8 @@ const Layout: FC<PropsWithChildren> = ({ children }) => {
   const showNav =
     !!user &&
     !path.startsWith('/site-details') &&
-    !path.startsWith('/inverters');
+    !path.startsWith('/inverters') &&
+    !path.startsWith('/link');
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


lib/components/Layout.tsx
```diff
@@ -12,7 +12,8 @@ const Layout: FC<PropsWithChildren> = ({ children }) => {
   const showNav =
     !!user &&
     !path.startsWith('/site-details') &&
-    !path.startsWith('/inverters');
+    !path.startsWith('/inverters') &&
+    !path.startsWith('/link');
```
ericcccsliu, line 16: @todo: figure out a more elegant way to do this perhaps

lib/components/form/LinkInverters.tsx
```diff
@@ -33,7 +34,7 @@ const SupportedInverters = () => {
             <img
```
ericcccsliu, line 34: do we want to replace this with Image? either way i think i got this in main too with last pr lol

ericcccsliu (approved):

lib/components/form/LinkInverters.tsx
AndrewLester: I was trying before pushing this and I think it'll require another div and just a bit more junk, can do after showcase tomorrow

PR #292, opened 2023-05-01
Improve centered moving average algorithm

amit-sawhney:
## Description

Improves the centered moving average algorithm by utilizing prefixSums to decrease time complexity from `O(n*p)` where `p` is the period size to `O(n)`

### DIFF
--- a/lib/graphs.ts
+++ b/lib/graphs.ts
@@ -201,8 +201,8 @@ export const calculateCenteredMovingAverage = (
   );
 
   // Calculate the centered moving average for each point
-  for (let i = 0; i < averagedPoints.length; i++) {
-    if (i == 0 || i == averagedPoints.length - 1) {
+  for (let i = 0; i < points.length; i++) {
+    if (i == 0 || i == points.length - 1) {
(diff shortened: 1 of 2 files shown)


lib/graphs.ts
```diff
@@ -171,49 +171,72 @@ export const getTrendAfterIndex = (
...
+    [0]
+  );
 
-    const range = Math.min(valsBefore, valsAfter, upperBound);
-    let currSum = 0;
-    for (let j = i - range; j < i + range + 1; j++) {
-      currSum += array[j].generation_kw;
+  // Calculate the centered moving average for each point
+  for (let i = 0; i < averagedPoints.length; i++) {
+    if (i == 0 || i == averagedPoints.length - 1) {
```
AndrewLester: Is this supposed to be this? It's returning empty rn
```suggestion
  for (let i = 0; i < points.length; i++) {
    if (i == 0 || i == points.length - 1) {
```

AndrewLester (approved):
Missing a proof but I'll let it go

PR #310, opened 2023-05-21
Fix site details search underlay

### LINKED ISSUE #309: Overlay behind search bar
**Describe the bug**
The location search bar in `site-details` endpoint has a black edges

**To Reproduce**

1. Go to `site-details`
2. See error

**Expected behavior**
No black overlay 

**Additional context**
image

anyaparekh:
## Status:
:rocket: Ready

## Description
No more black background in search bar!

Fixes #309 

## Screenshots
image

### DIFF
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -26,7 +26,7 @@
 
 /* Apply additional styles to .mapboxgl-ctrl-geocoder */
 .mapboxgl-ctrl-geocoder {
-  @apply h-14 min-w-full rounded-lg bg-ocf-black !important;
+  @apply h-14 min-w-full bg-transparent !important;
 }
 
 /* Apply styles to geocoder position in map */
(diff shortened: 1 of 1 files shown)


styles/globals.css
```diff
@@ -26,7 +26,7 @@
 
 /* Apply additional styles to .mapboxgl-ctrl-geocoder */
 .mapboxgl-ctrl-geocoder {
-  @apply h-14 min-w-full bg-ocf-black !important;
+  @apply h-14 min-w-full rounded-lg bg-ocf-black !important;
```
ericcccsliu: I'm pretty sure this class represents the wrapper that contains the geocoder—my concern rn is that making this rounded might cause problems in the future if, for some reason, the geocoder itself needs to be changed. The background color is probably what's causing the black underlay—maybe changing that instead of changing the container to fit its contents would be better

```suggestion
  @apply h-14 min-w-full bg-transparent !important;
```

ericcccsliu (approved):

styles/globals.css
anyaparekh: yep this makes sense!

PR #311, opened 2023-05-22
Move "show nav" checks from Layout to individual pages

### LINKED ISSUE #288: Move "show nav" checks from Layout to individual pages
We currently check whether or not to show the navigation UI here, in the Layout:
https://github.com/openclimatefix/pv-sites-mobile/blob/7c5631672e77673e60073d76b6718d251ac1896d/lib/components/Layout.tsx#L12-L16

For context, the navigation UI is hidden on these screens to promote the form's flow, rather than escaping to the rest of the app.

The way we do that is by checking if the current URL is one of the URLs for pages we want to show the nav on. This solution is a bit hacky, mostly since we now have to remember to change these strings if we change URL path names.

An alternative solution is to determine this "show nav", or "hide nav" (which we'll go with since it's fewer pages), in the page tsx files directly. The way this can be done is by attaching a function to each page's page component which returns true if the nav should be hidden, and is assumed to be false if the function doesn't exist. Here's a small example that actually renders an entire layout for each individual page, but we can just pass this variable into the `Layout` component (inside `_app.tsx`) to show/hide the nav.

https://nextjs.org/docs/pages/building-your-application/routing/pages-and-layouts#per-page-layouts

neha-vard:
## Description
Determine whether to hide nav bar in the page tsx files directly.

Fixes #288

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -12,7 +12,6 @@ import duration from 'dayjs/plugin/duration';
 import dayjs from 'dayjs';
 import { Site } from '~/lib/types';
 import { AppProvider } from '~/lib/provider';
-import { NextComponentType, NextPageContext } from 'next';
 
 dayjs.extend(utc);
 dayjs.extend(timezone);
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -12,6 +12,7 @@ import duration from 'dayjs/plugin/duration';
 import dayjs from 'dayjs';
 import { Site } from '~/lib/types';
 import { AppProvider } from '~/lib/provider';
+import { NextComponentType, NextPageContext } from 'next';
 
 dayjs.extend(utc);
 dayjs.extend(timezone);
@@ -25,6 +26,7 @@ const App: AppType<AppProps> = ({ Component, pageProps }) => {
       site_list: pageProps.sites ?? [],
     },
   };
+  const hideNav = (Component as any).hideNav || false;
 
   return (
     <UserProvider user={pageProps.user}>
@@ -48,7 +50,7 @@ const App: AppType<AppProps> = ({ Component, pageProps }) => {
             />
             <meta name="theme-color" content="#14120E" />
           </Head>
-          <Layout>
+          <Layout hideNav={hideNav}>
             <Component {...pageProps} />
           </Layout>
         </AppProvider>
--- a/pages/inverters/[uuid].tsx
+++ b/pages/inverters/[uuid].tsx
@@ -56,4 +56,4 @@ export const getServerSideProps = withSites({
 
 export default Inverters;
 
-(Inverters as any).hideNav = true;
+Inverters.hideNav = true;
--- a/pages/inverters/[uuid].tsx
+++ b/pages/inverters/[uuid].tsx
@@ -55,3 +55,5 @@ export const getServerSideProps = withSites({
 });
 
 export default Inverters;
+
+(Inverters as any).hideNav = true;
--- a/pages/site-details/[uuid].tsx
+++ b/pages/site-details/[uuid].tsx
@@ -26,3 +26,5 @@ export const getServerSideProps = withSites({
 });
 
 export default NewSiteDetails;
+
+(NewSiteDetails as any).hideNav = true;
--- a/pages/site-details/[uuid].tsx
+++ b/pages/site-details/[uuid].tsx
@@ -8,7 +8,7 @@ interface NewSiteDetailsProps {
   sites: Site[];
 }
 
-const NewSiteDetails: FC<NewSiteDetailsProps> = ({ sites }) => {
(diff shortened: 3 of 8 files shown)


pages/inverters/[uuid].tsx
```diff
@@ -55,3 +55,5 @@ export const getServerSideProps = withSites({
 });
 
 export default Inverters;
+
+(Inverters as any).hideNav = true;
```
Comment (AndrewLester) [inline pages/inverters/[uuid].tsx]: These casts to `any` (in page tsx files) don't seem to be necessary for building/linting the app on my setup at least. Were you running into a problem that required adding them? I think they can be removed otherwise.

It is a bit strange that you can get away without a cast to `any` in a case like this, but I think it's actually a builtin TypeScript case specifically for React that allows for this.

AndrewLester (approved):
Great work, this is exactly what we needed here.

ericcccsliu (approved):
only thing to note is the casting to `any` (which Andrew mentioned). Great work!

pages/_app.tsx
```diff
@@ -25,6 +25,7 @@ const App: AppType<AppProps> = ({ Component, pageProps }) => {
       site_list: pageProps.sites ?? [],
     },
   };
+  const hideNav = (Component as any).hideNav || false;
```
ericcccsliu, line 28: We can remove this instance of casting to `any` by adding an optional `hideNav` prop to `Component`. [This](https://stackoverflow.com/questions/70644647/how-can-i-add-a-new-property-to-component-in-app-tsx-in-next-js-with-typescript#:~:text=You%20need%20to%20extend%20AppProp%20and%20add%20your%20custom%20attributes) stackoverflow thread might be helpful!

pages/site-details/index.tsx
```diff
@@ -8,3 +8,5 @@ const NewSiteDetails: FC = () => {
 
 export default NewSiteDetails;
 export const getServerSideProps = withSites();
+
+(NewSiteDetails as any).hideNav = true;
```
ericcccsliu: `NewSiteDetails` doesn't seem to need to be a `FC` here, maybe removing that type will allow us to remove this cast to `any`

pages/site-details/[uuid].tsx
```diff
@@ -26,3 +26,5 @@ export const getServerSideProps = withSites({
 });
 
 export default NewSiteDetails;
+
+(NewSiteDetails as any).hideNav = true;
```
Comment (ericcccsliu) [inline pages/site-details/[uuid].tsx]: so there's three ways to handle this case I think:

1. We can leave it the same, which would be easy, wouldn't increase the amount of code we have, and would still be clear enough. Casting things to `any` is really not a best practice when using TypeScript, though, so that would be the tradeoff. 
2. We can add an optional `hideNav` prop to `NewSiteDetailsProps`
3. We can use [`Object.assign()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) to [implicitly add](https://stackoverflow.com/questions/57712682/react-functional-component-static-property#:~:text=With%20implicit%20typing%20(Best%20Solution)) the `hideNav` property (and its inferred boolean type) to our component. 

@AndrewLester @neha-vard what do y'all think would be best? I think the third option is better than the second, but I don't have a strong preference for the third over the first.

ericcccsliu (approved):

pages/site-details/[uuid].tsx
Comment (AndrewLester) [reply pages/site-details/[uuid].tsx]: Another option, that might be a little simpler, is just to remove the `FC` type on the variable `NewSiteDetails` and move the props type to the `{ sites }` parameter itself, so like typing the component without `FC`. This will cause `NewSiteDetails` to be typed as a standard function which allows for arbitrary property assignment.

pages/site-details/index.tsx
neha-vard: This worked!

pages/site-details/[uuid].tsx
Comment (neha-vard) [reply pages/site-details/[uuid].tsx]: @AndrewLester's suggestion worked and seems to align with @ericcccsliu's suggestion for site-details/index.tsx

pages/_app.tsx
neha-vard: image

This is actually something I tried before using as any, but I'm getting this Typescript error: 
Property 'hideNav' does not exist on type 'NextComponentType<NextPageContext, any, any>'.
Property 'hideNav' does not exist on type 'ComponentClass<any, any> & { getInitialProps?(context: NextPageContext): any; }'.ts(2339)

AndrewLester: At the end of the day the only type safety we're doing here is ensuring there's no typo in `hideNav` on line 28 specifically, since this won't be recognized by the page files that actually _set_ `hideNav`. For this reason, the `any` here doesn't seem too big of a deal, and it'll take some more annoying typing to get this to check without `any`.

neha-vard: @AndrewLester That makes sense, let me know if I should merge!

PR #313, opened 2023-05-25
Small fixes

AndrewLester:
- Deployment
- Deployment
- Deployment
- Make next error text white
- Change no site redirect
- Remove inverters fetch

## Status:

:rocket: Ready

## Description

Fixes #<number>

### DIFF
--- a/lib/sites.ts
+++ b/lib/sites.ts
@@ -30,7 +30,8 @@ export function useSites() {
     `${process.env.NEXT_PUBLIC_API_BASE_URL_GET}/sites`
   );
 
-  return { sites: sites?.site_list ?? [], error, isLoading };
+  // TODO: Paginate this globally somehow. Right now there is too much site data being fetched on the aggregate dashboard
+  return { sites: sites?.site_list?.slice(0, 5) ?? [], error, isLoading };
 }
 
 /**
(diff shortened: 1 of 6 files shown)


lib/sites.ts
```diff
@@ -30,7 +30,8 @@ export function useSites() {
     `${process.env.NEXT_PUBLIC_API_BASE_URL_GET}/sites`
   );
 
-  return { sites: sites?.site_list ?? [], error, isLoading };
+  // TODO: Paginate this globally somehow. Right now there is too much site data being fetched on the aggregate dashboard
+  return { sites: sites?.site_list?.slice(0, 5) ?? [], error, isLoading };
```
ericcccsliu, line 34: this probably won't be too much of an issue with individual homeowners  right? but yeah I think we should make this into an issue so that this problem is easier to find.

pages/index.tsx
```diff
@@ -6,7 +6,7 @@ export default function Index() {
 
 export const getServerSideProps = withSites({
   async getServerSideProps({ sites }) {
-    const destination = sites.length === 0 ? '/form/location' : '/dashboard';
+    const destination = sites.length === 0 ? '/site-details' : '/dashboard';
```
ericcccsliu, line 9: oh, nice catch

ericcccsliu (approved):

